PR #3, opened 2018-02-18
Draft: Event Details Column

bordoni:
**Video Demo:** http://i.bordoni.me/pc8F

---

**Extra UI done here:**
screen recording 2018-02-18 at 19 16
screen recording 2018-02-18 at 19 16


src/Tribe/Blocks/Event_Details.php
```diff
@@ -23,7 +23,7 @@ public function name() {
 	 * @return string
 	 */
 	public function render( $attributes = array() ) {
-		return 'My Own test';
+		return 'Event Details';
```
borkweb: Are we avoiding `__()` on purpose? Asking because I don't actually know if Gutenberg handles localization differently.

src/modules/blocks/style.pcss
```diff
@@ -0,0 +1,157 @@
...
+	background: #fafafa;
+	border: 1px solid #eee;
+	margin: 30px 0;
+	zoom: 1;
+
+	&::before,
+	&::after {
+		content: '';
+		display: table;
+		clear: both;
```
borkweb: Alphabetize!

src/modules/elements/datetime.js
```diff
@@ -0,0 +1,125 @@
...
+		if ( 'string' === typeof date ) {
+			date = moment( date );
+		}
+
+		return date.minutes( round15( date.minutes() ) ).format( format );
+	}
+
+	onChange( date ) {
+		// Start defaults to now
+		var dateMoment = moment();
```
borkweb: Use `let` instead of `var` here.

src/modules/blocks/event-details/organizers.js
```diff
@@ -0,0 +1,232 @@
...
+					onClick={ () => removeOrganizer( organizer.id ) }
+				/>
+			</li>
+		);
+	}
+
+	render() {
+		const { focus, addOrganizer } = this.props;
+		const organizers = this.getOrganizers();
+		var list = (
```
borkweb: `let` rather than `var`

src/modules/elements/datetime.js
```diff
@@ -0,0 +1,125 @@
...
+						aria-expanded={ isOpen }
+					>
+						{ this.formatDate( this.props.datetime, true ) }
+					</button>
+				) }
+				renderContent={ () => <DateTimePicker key="datetime-picker" currentDate={ this.props.datetime } onChange={ this.onChange } /> }
+			/>
+		);
+
+		const label = this.props.label ? <strong className='tribe-detail-label'>{ this.props.label } </strong> : '';
```
nealfennimore: Might want to return `null` or `false` here to not render anything, otherwise it's going to add an empty `<!-- react-text -->` comment in

src/modules/elements/meta-group.js
```diff
@@ -0,0 +1,38 @@
...
+ * Internal dependencies
+ */
+
+/**
+ * Module Code
+ */
+class MetaGroup extends Component {
+	static defaultProps = { className: '', children: null }
+
+	constructor() {
```
nealfennimore: I don't think you need this constructor here, since you're not doing any other work inside of it. I haven't used the `static` properties though yet, so I can't say if that's the case

src/modules/elements/search-posts.js
```diff
@@ -0,0 +1,164 @@
...
+
+
+	/**
+	 * Returns an event handler triggered when hovering a block.
+	 *
+	 * @param   {Object} block Block object.
+	 * @return  {Func}         Event Handler.
+	 */
+	createToggleBlockHover( block ) {
+		if ( ! this.props.onHover ) {
```
nealfennimore: You can add `defaultProps` to `SearchPosts` with a noop function being `onHover` if you want to avoid this statement here

src/modules/elements/terms-list.js
```diff
@@ -0,0 +1,142 @@
...
+		return (
+			<ul className={ this.getTermListClassName( 0 ) }>
+				{ terms.map( ( term, index ) => this.renderTermListItem( term, index + 1 === terms.length, 0 ) ) }
+			</ul>
+		);
+	}
+
+	renderTermListItem( term, isLast, level ) {
+		const childTerms = this.getTerms( term.id );
+		const separator = ! isLast ? <span>{ this.props.termSeparator } </span> : '';
```
nealfennimore: Might want to return null or false here too

src/modules/elements/organizer-form.js
```diff
@@ -0,0 +1,170 @@
...
+					key='tribe-organizer-form'
+				>
+					<Placeholder key="placeholder">
+						<Spinner />
+					</Placeholder>
+				</div>
+			];
+		}
+
+		return [
```
nealfennimore: Nice didn't realize you could return arrays now

src/Tribe/Blocks/Event_Details.php
bordoni: Didn't tackle this, it's just a place holder.

src/modules/elements/organizer-form.js
bordoni: Yeah, but all first level elements need a `key` attr, otherwise it will complain

PR #41, opened 2018-03-31
Add new <Input /> element

mitogh:
Ticket: https://central.tri.be/issues/102554

Provides more flexibility for reuse and custom validations

### DIFF
--- a/src/modules/blocks/style.pcss
+++ b/src/modules/blocks/style.pcss
@@ -443,6 +443,15 @@
 	}
 }
 
+input.is-invalid {
+	border: 1px solid #e62f2f;
+}
+
+input:focus.is-invalid {
+	border: 1px solid #e62f2f;
+	outline: none;
+}
+
 /**
  * Reset our Width of the Gutenberg Editor
  */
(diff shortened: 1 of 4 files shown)


src/modules/elements/input.js
```diff
@@ -0,0 +1,105 @@
+const {Component} = wp.element
```
bordoni: Lets use  import `@wordpress/element` like we do in other places.

src/modules/elements/input.js
```diff
@@ -0,0 +1,105 @@
...
+import PropTypes from 'prop-types'
```
bordoni: Would love to see us keep the same doc blocks across the files for where things come from, open to discussion on if they are really valid.

src/modules/blocks/style.pcss
```diff
@@ -443,6 +443,15 @@
 	}
 }
 
+input.is-invalid {
+	border: 1px solid #e62f2f;
+}
+
+input:focus.is-invalid {
```
bordoni: Lets use indentation to avoid repetition and we should have this as a name-spaced class.

mitogh:
Ready for another set of eyes.

mitogh:
Fixed based on eslint rules

PR #46, opened 2018-04-05
Prevent Google Maps JS API to re-render every time

bordoni:


src/modules/elements/google-map.js
```diff
@@ -168,21 +167,37 @@ class GoogleMap extends Component {
...
-			mapTypeId: 'roadmap'
+			mapTypeId: mapType
 		}
 
-		this.map = new maps.Map( node, mapConfig );
-		const marker = new google.maps.Marker( {
-			position: location,
-			map: this.map,
-		} );
+		let map;
```
mitogh: I think you can simplify the next conditional block like: 

```js
let { map } = this.state;

if ( ! map ) {
  const mapRef = this.refs.map;
  ...
}
````

Comment (mitogh) [reply src/modules/elements/google-map.js]: So you don't need the else block

File: src/modules/elements/google-map.js [hunk 2]
```diff
@@ -168,21 +167,37 @@ class GoogleMap extends Component {
...
+			this.setState( { marker: marker } )
+		}
 	}
 
 	renderInteractive() {
 		return (
 			<div
 				className='tribe-editor-map__element'
-				ref="map"
+				ref='map'
```
mitogh: I think using `strings` for reference is consider legacy code and is going to be deprecated at some point.

> If you worked with React before, you might be familiar with an older API where the ref attribute is a string, like "textInput", and the DOM node is accessed as this.refs.textInput. We advise against it because string refs have some issues, are considered legacy, and are likely to be removed in one of the future releases.

[Source](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs)

PR #47, opened 2018-04-10
Add select for Countries and states

mitogh:


src/utils/geo-data.js
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+/**
+ * Return an object from the global namespace or empty object if does not exist
+ *
+ * @param name
+ * @returns {*|{}}
+ */
+export function getObject( name ) {
```
bordoni: We should maybe move this into a more global Utils file, it's not specific to `geo-data` right?

src/utils/geo-data.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Get name from a container object based on the code of the element otherwise an empty string.
+ *
+ * @param code
+ * @param container
+ * @returns {*|string}
+ */
+export function getName( code, container ) {
```
bordoni: Should we use the lodash `get` instead of this?

mitogh:
Latest changes applied @bordoni 🎉

bordoni:
Add the changelog and merge.

PR #50, opened 2018-04-16
Add store implementations for Venues and Organizer

mitogh:
With this change the data is stored in a store instead of the component as the components can be unmounted or removed from the virtual DOM,
also the stores are the ones in charge to perfom the requests for the next pages.

### DIFF
--- a/src/modules/blocks/event-details/organizers.js
+++ b/src/modules/blocks/event-details/organizers.js
@@ -28,7 +28,7 @@ import {
 	OrganizerForm,
 	SearchPosts,
 } from 'elements';
-import { store, STORE_NAME } from './../../data/organizers';
+import { store, STORE_NAME } from 'data/organizers';
 
 function CreateDropdown( { ...props } ) {
 	const { focus, addOrganizer } = props;
--- a/src/modules/blocks/event-details/organizers.js
+++ b/src/modules/blocks/event-details/organizers.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { unescape, union, uniqueId } from 'lodash';
+import { unescape, union, uniqueId, noop } from 'lodash';
 import { stringify } from 'querystringify';
 import classNames from 'classnames';
 
@@ -9,7 +9,7 @@ import classNames from 'classnames';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { withSelect } from '@wordpress/data'
+import { withSelect, select } from '@wordpress/data'
 import { Component, compose } from '@wordpress/element';
 
 import {
@@ -27,7 +27,8 @@ import {
 import {
 	OrganizerForm,
 	SearchPosts,
-} from 'elements'
+} from 'elements';
+import { store, STORE_NAME } from './../../data/organizers';
 
 function CreateDropdown( { ...props } ) {
 	const { focus, addOrganizer } = props;
@@ -102,7 +103,27 @@ class EventOrganizers extends Component {
 
 		this.state = {
 			overOrganizer: null,
+			organizers: [],
+			isLoading: false,
 		}
+		this.unsubscribe = noop;
+	}
+
+	componentDidMount() {
+		select( STORE_NAME ).fetch();
+		this.unsubscribe = store.subscribe( () => {
+			const { posts } = store.getState();
+			this.setState({ posts });
+		});
+	}
+
+	componentDidUpdate(prevProps, prevState) {
+		const { page } = store.getState();
+		select( STORE_NAME ).fetch({ page });
+	}
+
+	componentWillUnmount() {
+		this.unsubscribe();
 	}
 
 	renderOrganizerName( organizer ) {
@@ -111,23 +132,6 @@ class EventOrganizers extends Component {
 		}
 
 		return unescape( organizer.title.rendered ).trim();
-	}
-
-	getOrganizers( parentId = null ) {
-		if ( ! this.props.organizers ) {
-			return [];
-		}
-
-		const organizers = this.props.organizers.data;
-		if ( ! organizers || ! organizers.length ) {
-			return [];
-		}
-
-		if ( parentId === null ) {
-			return organizers;
-		}
-
-		return organizers.filter( organizer => organizer.parent === parentId );
 	}
 
 	renderOrganizerList() {
@@ -148,7 +152,6 @@ class EventOrganizers extends Component {
 			'tribe-current': focus && current
 		}
 
-		let renderButtons = null;
 		return (
 			<li
 				className={ classNames( classes ) }
@@ -168,7 +171,7 @@ class EventOrganizers extends Component {
 
 	render() {
 		const { focus, addOrganizer } = this.props;
-		const organizers = this.getOrganizers();
+		const { organizers, isLoading } = this.state
 		const hasOrganizers = 0 !== organizers.length
 		let list = null;
 		let actions = (
@@ -179,6 +182,8 @@ class EventOrganizers extends Component {
 					metaKey='_EventOrganizerID'
 					searchLabel={ __( 'Search for an organizer', 'the-events-calendar' ) }
 					iconLabel={ __( 'Add existing Organizer', 'the-events-calendar' ) }
+					store={store}
+					storeName={STORE_NAME}
 					focus={ hasOrganizers ? focus : true }
 					onSelectItem={ addOrganizer }
 				/>
@@ -190,7 +195,7 @@ class EventOrganizers extends Component {
 			</div>
 		)
 
-		if ( this.props.organizers.isLoading ) {
+		if ( isLoading ) {
 			list = (
 				<Placeholder style={{ minHeight: 50 }} key="placeholder">
 					<Spinner />
@@ -229,29 +234,6 @@ const applySelect = withSelect( ( select, props ) => {
 	}
 } );
 
-const applyWithAPIData = withAPIData( ( props ) => {
-	const { organizers } = props
-	let query = {
-		per_page: 100,
-		orderby: 'modified',
-		status: [ 'draft', 'publish' ],
-		order: 'desc',
-		include: organizers,
-	};
-
-	if ( ! organizers || 0 === organizers.length ) {
-		return {
-			organizers: [],
-		}
-	}
-
-	return {
-		organizers: `/wp/v2/tribe_organizer?${ stringify( query ) }`,
-	};
-} );
-
-
 export default compose(
 	applySelect,
-	applyWithAPIData,
 )( EventOrganizers );
--- a/src/modules/blocks/event-venue/venue.js
+++ b/src/modules/blocks/event-venue/venue.js
@@ -10,6 +10,7 @@ import classNames from 'classnames';
  */
 import { __ } from '@wordpress/i18n';
 import { Component, compose } from '@wordpress/element';
+import { store, STORE_NAME } from './../../data/venues';
 
 import {
 	Dropdown,
@@ -97,6 +98,9 @@ class VenueDetails extends Component {
 	constructor( props ) {
 		super( ...arguments );
 
+		this.state = {
+			isLoading: false,
+		}
 		this.renderVenueName = this.renderVenueName.bind( this )
 		this.renderVenue = this.renderVenue.bind( this )
 		this.renderActions = this.renderActions.bind( this )
@@ -199,6 +203,9 @@ class VenueDetails extends Component {
 					iconLabel={ __( 'Add existing venue', 'the-events-calendar' ) }
 					focus={ focus }
 					onSelectItem={ addVenue }
+					store={store}
+					storeName={STORE_NAME}
+					searchable
 				/>
 				<CreateDropdown
 					key='venue-create-dropdown'
@@ -218,7 +225,8 @@ class VenueDetails extends Component {
 	}
 
 	render() {
-		const { focus, addVenue, removeVenue, venue, isLoading } = this.props;
+		const { venue } = this.props;
+		const { isLoading } = this.state;
 
 		if ( isLoading ) {
 			return (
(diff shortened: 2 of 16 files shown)


src/modules/elements/search-posts/element.js
```diff
@@ -31,52 +31,94 @@ import './style.pcss';
 class SearchPosts extends Component {
 	static defaultProps = {
 		onHover: () => {},
```
nealfennimore: Maybe update this to use `noop` instead?

src/modules/data/organizers.js
```diff
@@ -0,0 +1,131 @@
...
+				dispatch( STORE_NAME ).setPage( page + 1 );
+				dispatch( STORE_NAME ).setTotal( totalPages );
+				dispatch( STORE_NAME ).addPosts( body );
+				dispatch( STORE_NAME ).unblock();
+			});
+		},
+	},
+} );
+
+export {
```
nealfennimore: You can also export the `const` above instead of having to export an object. While this does clear up what's allowed as an export in a single place, our eslint config should show you if something was not exported or not (as well as autocomplete if IDE supports it).

For example:
`export const STORE_NAME = 'tec.organizers';`
`export const store = registerStore( STORE_NAME, {`

If you wanted to export as an object, you could also do `import * as organizers from 'modules/data/organizers'`, and then you could you use it like: `organizers.store` or `organizers.STORE_NAME`.

You could also individually import the above like, `import {store, STORE_NAME} from 'modules/data/organizers'`

Nothing wrong with your approach here, just an fyi

src/modules/editor/utils/request.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Extract the heeaders from an XHR request object.
+ *
+ * @param xhr Ajax object
+ * @returns object with key / value pairs: key is the name of the header and the value represents the header value
+ */
+export function getResponseHeaders( xhr ) {
```
nealfennimore: Wow! All this to get some headers - nice work. 

Though I'm wondering why that `wp` object isn't returning the headers back to you in a more usable format ¯\_(ツ)_/¯

src/modules/editor/utils/request.js
```diff
@@ -0,0 +1,26 @@
...
+	return xhr.getAllResponseHeaders().trim()
+		// Split the headers on Carry Return  or New Line
+		.split( '\u000d\u000a' )
+		// Create an array of key/value pais separated by: ; or space
+		.map( ( entry ) => entry.split( '\u003a\u0020' ) )
+		// Crate an array of objects with key / value pairs
+		.map( (values) => {
```
nealfennimore: I think there is an issue with using `map` here, you'll be getting `undefined` elements in your array, since you're not returning a new `header` in all cases.

Might want want to switch it out for `filter` and check for values length there, or simply make it reduce:
```js
reduce((header, values)=>{
    if ( values.length === 2 ) {
        const [ key, value ] = values;
        header[ key ] = value;
    }
    return header;
}, {})
```
With the above you could also get rid of the next `reduce` in the sequence

src/modules/elements/terms-list/element.js
```diff
@@ -135,13 +135,17 @@ const applySelect = withSelect( ( select, props ) => {
 
 const applyWithAPIData = withAPIData( ( props ) => {
 	const { slug, terms } = props;
-	const query = stringify( {
+	const args = {
 		per_page: 100,
 		orderby: 'count',
 		order: 'desc',
-		include: terms,
+		// include: terms.length ? terms : false,
```
nealfennimore: Still need this? :)

src/modules/elements/search-posts/element.js
```diff
@@ -1,14 +1,14 @@
 /**
```
nealfennimore: I'm going to come back to this file for another look later

src/modules/data/organizers.js
mitogh: Amazing thanks for the feedback will follow the inline export I think makes things more easier to spot at first glance.

src/modules/editor/utils/request.js
mitogh: Problem is `apiRequest` from WP is only an abstraction on top of the jQuery ajax function so it does not return a Promise like `fetch` I was going to use `axios` or the polyfill for `fetch` but for now to avoid adding more dependency I think just having a function to get the headers from the XHR object makes more sense.

Let me know what do you think about this.

src/modules/editor/utils/request.js
mitogh: Fantastic! Will update accordingly.

src/modules/elements/terms-list/element.js
mitogh: No will remove it.

src/modules/elements/search-posts/element.js
```diff
@@ -30,53 +30,95 @@ import './style.pcss';
...
-		if ( ! this.props.items ) {
-			return [];
+		if ( store.getState ) {
+			this.saveState();
 		}
 
-		const items = this.props.items.data;
-		if ( ! items || ! items.length ) {
-			return [];
+		this.unsubscribe = store.subscribe(this.saveState)
```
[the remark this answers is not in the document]
mitogh: I'm removing the conditional instead as the is not required, because the initial state of the component should be the initial state of the store.

src/modules/blocks/event-venue/venue.js
```diff
@@ -10,6 +10,7 @@ import classNames from 'classnames';
  */
 import { __ } from '@wordpress/i18n';
 import { Component, compose } from '@wordpress/element';
+import { store, STORE_NAME } from './../../data/venues';
```
bordoni: We might want to avoid using paths, lets go with using the module names. If we need to create a Data Module it's fine.

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
-							id={ `editor-inserter__${ instanceId }` }
-							type='search'
-							placeholder={ searchLabel }
-							className='editor-inserter__search'
-							onChange={ ( event ) => console.log( event.target.value ) }
-						/>
-					 */}
-				<div role="menu" className={ classNames( 'tribe-element-search-posts-results' ) }>
-					{ this.renderList() }
+			<div className={classNames( 'tribe-element-search-posts' )} onScroll={this.onScroll}>
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+				{this.renderSearchInput()}
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+				<div role="menu" className={classNames( 'tribe-element-search-posts-results' )}>
+					{this.renderList()}
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+		const { searchable, searchLabel } = this.props;
+		const instanceId = uniqueId( 'search-' );
+
+		if ( ! searchable ) {
+			return null;
+		}
+
+		return <div>
+			<label htmlFor={`editor-inserter__${ instanceId }`} className="screen-reader-text">
+				{searchLabel}
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+	renderSearchInput() {
+		const { searchable, searchLabel } = this.props;
+		const instanceId = uniqueId( 'search-' );
+
+		if ( ! searchable ) {
+			return null;
+		}
+
+		return <div>
+			<label htmlFor={`editor-inserter__${ instanceId }`} className="screen-reader-text">
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+				{searchLabel}
+			</label>
+			<input
+				id={`editor-inserter__${ instanceId }`}
+				type='search'
+				placeholder={searchLabel}
+				value={this.state.search}
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -101,37 +140,44 @@ class SearchPosts extends Component {
...
+		if ( ! searchable ) {
+			return null;
+		}
+
+		return <div>
+			<label htmlFor={`editor-inserter__${ instanceId }`} className="screen-reader-text">
+				{searchLabel}
+			</label>
+			<input
+				id={`editor-inserter__${ instanceId }`}
```
bordoni: Spacing around `{}`

src/modules/elements/search-posts/element.js
```diff
@@ -149,7 +195,7 @@ class SearchPosts extends Component {
 	}
 
 	render () {
-		const { items, focus } = this.props;
+		const { focus  } = this.props;
```
bordoni: extra space here.

src/modules/elements/search-posts/element.js
```diff
@@ -30,53 +30,92 @@ import './style.pcss';
...
+		select( storeName ).fetch({
+			search: value,
+			orderBy: value ? 'relevance' : 'title',
+		});
 	}
 
-	renderList() {
-		const { items, focus } = this.props;
-		const isLoading = items.isLoading;
+	onScroll = (event) => {
```
bordoni: Spacing around `()`

src/modules/editor/utils/request.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function getResponseHeaders( xhr ) {
+	return xhr.getAllResponseHeaders().trim()
+		// Split the headers on Carry Return  or New Line
+		.split( '\u000d\u000a' )
+		// Create an array of key/value pais separated by: ; or space
+		.map( ( entry ) => entry.split( '\u003a\u0020' ) )
+		// Crate a single objects with key / value pairs
+		.reduce( (headers, values) => {
+			if ( values.length === 2 ) {
```
bordoni: Maybe invert?

src/modules/data/venues.js
```diff
@@ -0,0 +1,163 @@
...
+				page,
+			}, queryParams );
+
+			const search = query.search || '';
+			dispatch( STORE_NAME ).setSearch( search );
+			dispatch( STORE_NAME ).block();
+
+			apiRequest({
+				path: `/wp/v2/${ type }?${ stringify(query) }`
+			}).then((body, status, xhr) => {
```
bordoni: Spacing around `()`

src/modules/data/venues.js
```diff
@@ -0,0 +1,163 @@
...
+				order: 'asc',
+				page,
+			}, queryParams );
+
+			const search = query.search || '';
+			dispatch( STORE_NAME ).setSearch( search );
+			dispatch( STORE_NAME ).block();
+
+			apiRequest({
+				path: `/wp/v2/${ type }?${ stringify(query) }`
```
bordoni: Spacing around `()`

src/modules/data/venues.js
```diff
@@ -0,0 +1,163 @@
...
+					...state,
+					search: action.search.trim(),
+				}
+		}
+		return state;
+	},
+
+	actions: {
+		setSearch( search ) {
+				return {
```
bordoni: Weird indentation.

src/modules/blocks/event-details/organizers.js
```diff
@@ -27,7 +27,8 @@ import {
 import {
 	OrganizerForm,
 	SearchPosts,
-} from 'elements'
+} from 'elements';
+import { store, STORE_NAME } from './../../data/organizers';
```
bordoni: Use the module, instead of path.

src/modules/editor/utils/request.js
mitogh: I think as we are using `strict` comparison there are benefits or problems with false assignments.

src/modules/blocks/event-venue/venue.js
mitogh: Problem with doing so for this specifc case is most stores are going to be named `store` and we might need to creating an entry point with a larger complexity such as: 

```js
export { store as VenueStore, STORE_NAME as VenueStoreName } from './venues';;
export { store as OrganizerStore, STORE_NAME as OrganizerStoreName } from './organizers';
```

I think instead of doing so we might want to use an alias instead such as: 

```js
alias: {
  data: path.resolve(__dirname, 'src/modules/data')
}
```

So we can use as: 

```js
import { store, STORE_NAME } from 'data/venues';
````

let me know what you think @bordoni

mitogh:
Just a question about favor the alias instead of entry points.

src/modules/blocks/event-venue/venue.js
nealfennimore: Just to add some choices, you can either add a babel alias or a webpack alias:
https://www.npmjs.com/package/babel-plugin-module-resolver
https://webpack.js.org/configuration/resolve/

I'd recommend a webpack alias if everything is always being built with webpack

nealfennimore: Also, let me know if you go an alias route - I can hook up the linter to automatically ensure it's getting the right path from the alias

mitogh: Thanks for the notes @nealfennimore I was voting for: 

- https://webpack.js.org/configuration/resolve/

But let's wait for @bordoni feedback.

mitogh:
New alias has been added.

PR #51, opened 2018-04-16
Add new `<Dashboard>` component

mitogh:
Ticket: https://central.tri.be/issues/103738

Usage of this component:

```
<Dashboard open className="custom" direction={directions.up}>
 <AnyComponent></AnyComponent>
</Dashboard
```

### DIFF
--- a/src/modules/elements/dashboard/element.js
+++ b/src/modules/elements/dashboard/element.js
@@ -65,7 +65,7 @@ export class Dashboard extends Component {
 		this.setState({ open: this.props.open }, this.setupListeners );
 	}
 
-	componentWillUnMount() {
+	componentWillUnmount() {
 		this.removeListeners();
 	}
 
--- a/src/modules/elements/dashboard/element.js
+++ b/src/modules/elements/dashboard/element.js
@@ -3,6 +3,7 @@
  */
 import { trim } from 'lodash';
 import PropTypes from 'prop-types'
+import classNames from 'classnames';
 
 /**
  * Internal dependencies
@@ -35,6 +36,12 @@ export class Dashboard extends Component {
 		open: false,
 		className: '',
 		direction: directions.down,
+	}
+	
+	static propTypes = {
+		open: PropTypes.bool,
+		className: PropTypes.string,
+		direction: PropTypes.oneOf( Object.keys( directions ) ),
 	}
 
 	constructor() {
@@ -165,17 +172,12 @@ export class Dashboard extends Component {
 	getContainerClass() {
 		const { className, direction } = this.props;
 
-		const containerClass = [
+		return classNames(
 			'tribe-dashboard-container',
-			this.state.open ? 'tribe-dashboard-container--open' : '',
 			`tribe-dashboard-container--${direction}`,
-			...className.split(' ')
-		];
-
-		return containerClass
-			.map( trim )
-			.filter( item => item && item.length )
-			.join( ' ' );
+			{ 'tribe-dashboard-container--open': this.state.open },
+			...className
+		);
 	}
 
 	render() {
@@ -187,10 +189,4 @@ export class Dashboard extends Component {
 			</div>
 		);
 	}
-}
-
-Dashboard.propTypes = {
-	open: PropTypes.bool,
-	className: PropTypes.string,
-	direction: PropTypes.oneOf( Object.keys( directions ) ),
 }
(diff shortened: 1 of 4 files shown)


src/modules/elements/dashboard/element.js
```diff
@@ -0,0 +1,196 @@
...
+
+	/**
+	 * Construct a string with the appropiate classs for the main container of the component
+	 *
+	 * @returns {string}
+	 */
+	getContainerClass() {
+		const { className, direction } = this.props;
+
+		const containerClass = [
```
nealfennimore: Might want to use the `classnames` package here to clean this up. It should already be installed in this repo: https://github.com/moderntribe/events-gutenberg/blob/277cce8dbad071e21c955bcf766f8af444f2c23b/src/modules/blocks/event-details/block.js#L177

src/modules/elements/dashboard/element.js
```diff
@@ -0,0 +1,196 @@
...
+	isPartOfDashboard( node ) {
+		return searchParent( node, (testNode) => {
+			return testNode.classList.contains( 'tribe-dashboard-container' ) || testNode.classList.contains( 'tribe-dashboard' )
+		});
+	}
+
+	/**
+	 * Outside method to open the <Dashboard /> component using a reference
+	 */
+	open() {
```
nealfennimore: I'm not really sure where `open` and the `toggle` method are being utilized with the Dashboard or children ( I don't see them used anywhere )

src/modules/elements/dashboard/element.js
```diff
@@ -0,0 +1,196 @@
...
+			<div className={this.getContainerClass()}>
+				<div className="tribe-dashboard">
+					{this.props.children}
+				</div>
+			</div>
+		);
+	}
+}
+
+Dashboard.propTypes = {
```
nealfennimore: Can you move this where `defaultProps` is?

src/modules/elements/dashboard/element.js
```diff
@@ -0,0 +1,196 @@
...
+			this.close();
+		}
+	}
+
+	/**
+	 * Detect clicks on the document and test if they are part of the <Dashboard> component if not close the component.
+	 *
+	 * @param event
+	 */
+	onClickOutside = (event) => {
```
nealfennimore: There's a good plugin here if this becomes an issue elsewhere: https://github.com/Pomax/react-onclickoutside

src/modules/elements/dashboard/element.js
```diff
@@ -0,0 +1,196 @@
+/**
```
nealfennimore: I'm a little confused with how this works. Dashboard is supposed to show when it mounts, adds event listeners, but then hides `children` when clicked somewhere outside the Dashboard ( and event listeners are thus removed )?

It sort of looks like one of these patterns might work better for this component:
https://reactjs.org/docs/higher-order-components.html
https://reactjs.org/docs/render-props.html

src/modules/elements/dashboard/element.js
mitogh: At the moment `toggle` is not being used on the component is more an API so external components can trigger this behavior from outside of this component using a reference to this component.

At the moment `open` is only called from `toggle` but is also available via the reference.

In most cases: 

- `open()`
- `close)()`
- `toggle()`

Are not going to be used but I decided to add them in case a call to this action is required with a reference to the dashboard component.

In most cases the property `<Dashboard open>` on the component should be enough and with unmounting the component should be enough to remove any listeners attached to this component.

src/modules/elements/dashboard/element.js
mitogh: Thanks at the moment this component is not used anywhere but the idea is to render this component to display extra content as a POP UP for example.

So the event listeners are attached when the component is created and is opened .

Otherwise the events listeners are not attached, if the component is either closed or removed from the virtual DOM the events are going to be removed using the `componentWillUnMount` callback.

Let me know if you have any other questions about this.

src/modules/elements/dashboard/element.js
mitogh: Great thanks if we need this logic for more components I think would be better to use the dependency you are suggesting.

src/modules/elements/dashboard/element.js
```diff
@@ -65,6 +65,10 @@ export class Dashboard extends Component {
 		this.setState({ open: this.props.open }, this.setupListeners );
 	}
 
+	componentWillUnMount() {
```
nealfennimore: This have gotten me before too, but this should be `componentWillUnmount`

mitogh:
Ready for another set of eyes.

Thanks

PR #66, opened 2018-05-18
Prevent warning when using `moment` with no explicit format

mitogh:


src/modules/elements/date-picker/element.js
```diff
@@ -105,5 +108,3 @@ class DatePicker extends Component {
 		);
 	}
 }
-
-export default DatePicker;
```
bordoni: Why no longer export?

mitogh: Oh! great question; As the class is the only and the default part of this file being exported I moved into an inline export instead.

- https://github.com/moderntribe/events-gutenberg/blob/7eddf4417e11226045f3cf5f2eb019e39c613ec1/src/modules/elements/date-picker/element.js#L27

Let me know what do you think about that.

Thanks

bordoni: Ahh ok ;)

PR #69, opened 2018-05-22
Add Tags and Categories blocks

bordoni:
_Ref:_ [C#102644](http://central.tri.be/issues/102644), [C#102645](http://central.tri.be/issues/102645)
---

### DIFF
--- a/src/modules/blocks/event-category/block.js
+++ b/src/modules/blocks/event-category/block.js
@@ -1,24 +1,13 @@
 /**
  * External dependencies
  */
-import { noop, trim, isEmpty } from 'lodash';
 import classNames from 'classnames';
-import React from 'react';
 
 /**
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
-
-import {
-	Dropdown,
-	IconButton,
-	Dashicon,
-	ToggleControl,
-	TextControl,
-	PanelBody,
-} from '@wordpress/components';
 
 import {
 	InspectorControls,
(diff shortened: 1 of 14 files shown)


src/modules/blocks/event-category/block.js
```diff
@@ -0,0 +1,64 @@
...
+			<section key="event-category-box" className="tribe-editor-block">
+				<div className="tribe-editor__event-category">
+					{ this.renderList() }
+				</div>
+			</section>
+		);
+	}
+
+	renderList() {
+		const { attributes } = this.props;
```
mitogh: Seems like this one is not being used `attributes`

src/modules/blocks/event-tags/block.js
```diff
@@ -0,0 +1,64 @@
...
+			<section key="event-tags-box" className="tribe-editor-block">
+				<div className="tribe-editor__event-tags">
+					{ this.renderList() }
+				</div>
+			</section>
+		);
+	}
+
+	renderList() {
+		const { attributes } = this.props;
```
mitogh: Seems like this one isn't used either over here.

src/modules/elements/terms-list/element.js
```diff
@@ -88,31 +88,51 @@ class TaxonomiesElement extends Component {
 		);
 	}
 
-	render() {
-		const { attributes, focus, setAttributes, slug } = this.props;
-		const terms = this.getTerms();
+	renderEmpty() {
+		const { renderEmpty, slug } = this.props;
+		const key   = `tribe-terms-${ slug }`;
```
mitogh: Space after `key` until the `=` sign.

PR #72, opened 2018-05-25
Website block

bordoni:
_Ref:_ [C#102642](http://central.tri.be/issues/102642)
---


src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
+/**
+ * External dependencies
+ */
+import classNames from 'classnames';
+import { union, without, isEmpty, noop, pick } from 'lodash';
```
mitogh: Seems like: ` union, without, isEmpty` are not used on this one.

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
...
+			this.renderLabelInput( placeholder ),
+			this.renderUrlInput()
+		];
+	}
+
+	renderUrlInput() {
+		const { eventUrl, setAttributes } = this.state;
+		const { isSelected } = this.props;
+
+		const buttonLabel = eventUrl ? __( 'Edit Website', 'events-gutenberg' ) : __( 'Insert Website', 'events-gutenberg' );
```
mitogh: Seems like `buttonLabel` is not being used.

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
...
+		this.unsubscribe = noop;
+	}
+
+	componentDidMount() {
+		this.unsubscribe = store.subscribe( () => {
+			const state = store.getState();
+			this.setState( pick( state, VALID_PROPS ) );
+		} );
+
+		store.dispatch( {
```
mitogh: I don't think `label` or `eventUrl` are on the Details store.

I think we need to add a defautl over here: 

-  https://github.com/moderntribe/events-gutenberg/blob/7e16fbd6ed3e8e64c9c0279d351596136e35fa57/src/modules/data/details/index.js#L23-L34

On the default state

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
...
+					onChange={ ( nextContent ) => setAttributes( { eventUrl: nextContent } ) }
+					placeholder={ __( 'Website URL', 'events-gutenberg' ) }
+				/>
+			</div>
+		);
+	}
+
+	renderLabelInput( placeholder ) {
+		const { label, setAttributes } = this.state;
+		const { isSelected } = this.props;
```
mitogh: Seems like `isSelected` is not used on this method.

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
...
+
+		return (
+			<div key='tribe-events-website-label' className="tribe-events-website-link tribe-events-website">
+				<PlainText
+					id="tribe-events-website-link"
+					value={ label }
+					onChange={ ( nextContent ) => setAttributes( { label: nextContent } ) }
```
mitogh: If the fields are on the store we need to dispatch an action here instead with the value as the listener will set the attributes over here: 

- https://github.com/moderntribe/events-gutenberg/blob/7e16fbd6ed3e8e64c9c0279d351596136e35fa57/src/modules/blocks/event-website/index.js#L64

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,155 @@
...
+			return null;
+		}
+
+		return (
+			<div key='tribe-events-website-url' className="tribe-events-website-url">
+				<Dashicon icon="admin-links" />
+				<PlainText
+					id="tribe-events-website-link"
+					value={ eventUrl }
+					onChange={ ( nextContent ) => setAttributes( { eventUrl: nextContent } ) }
```
mitogh: Here we should dispatch an action if the value is going to be saved on the store for example: 

- https://github.com/moderntribe/events-gutenberg/blob/7e16fbd6ed3e8e64c9c0279d351596136e35fa57/src/modules/blocks/event-subtitle/block.js#L369-L374

As the listener on `index.js` is going to set the attributes when the store changes.

src/modules/blocks/event-website/style.pcss
```diff
@@ -0,0 +1,72 @@
...
+		background-color: transparent;
+		border: 1px solid #E1E3E6;
+		display: inline-flex;
+		font-family: 'Helvetica', 'sans-serif';
+		font-weight: bold;
+		justify-content: center;
+		margin-right: 15px;
+		min-height: 54px;
+		padding: 0 20px;
+		font-size: 1.2em;
```
mitogh: I think we should use `rem` units instead for fonts as `em` might change based on the context and the one being used on the global for `rem` is 16px. Which I think is consistent with the design.

src/modules/blocks/event-website/block.js
```diff
@@ -0,0 +1,163 @@
...
+
+	renderLabelInput( placeholder ) {
+		const { eventUrlLabel, setAttributes } = this.state;
+
+		return (
+			<div key='tribe-events-website-label' className="tribe-events-website-link tribe-events-website">
+				<PlainText
+					id="tribe-events-website-link"
+					value={ eventUrlLabel }
+					onChange={ ( nextContent ) => setAttributes( { eventUrlLabel: nextContent } ) }
```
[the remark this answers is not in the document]
mitogh: In that case you might want to remove it from `VALID_PROPS` as that one will be used to get the values from the store when the store changes and  `eventUrlLabel` is part of that array. 

So in some situations that value might be override by a store change and will be set to `undefined` if does not exist on the store.

bordoni:
Ready for another set of 👁 👁

PR #80, opened 2018-06-08
Simplest version of the Featured Image

bordoni:
_Ref:_ [C#103175](http://central.tri.be/issues/103175)
---

### DIFF
--- a/src/modules/blocks/featured-image/block.js
+++ b/src/modules/blocks/featured-image/block.js
@@ -18,8 +18,6 @@ import {
 /**
  * Internal dependencies
  */
-import { getSetting } from 'editor/settings';
-import './style.pcss';
 
 /**
  * Module Code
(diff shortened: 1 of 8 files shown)


src/modules/blocks/featured-image/block.js
```diff
@@ -0,0 +1,101 @@
...
+
+import {
+	InspectorControls,
+} from '@wordpress/editor';
+
+/**
+ * Internal dependencies
+ */
+import { getSetting } from 'editor/settings';
+import './style.pcss';
```
mitogh: Do we have a style for this one as doesn't seem be added in this PR?

src/modules/blocks/featured-image/block.js
```diff
@@ -0,0 +1,101 @@
...
+	const { getMedia, getPostType } = select( 'core' );
+	const { getEditedPostAttribute } = select( 'core/editor' );
+	const featuredImageId = getEditedPostAttribute( 'featured_media' );
+
+	return {
+		image: featuredImageId ? getMedia( featuredImageId ) : null,
+	};
+} );
+
+export default compose(
```
mitogh: I think in this case we don't plan to use more HOC so it might be easier to avoid `compose` and call it like: `export default applySelect( FeaturedImage );`

src/modules/blocks/featured-image/block.js
```diff
@@ -0,0 +1,101 @@
...
+		return [
+			this.renderUI(),
+		];
+	}
+
+	renderUI() {
+		return (
+			<section key="featured-image" className="tribe-editor__block">
+				<div className="tribe-editor__featured-image">
+					{ this.renderImage() }
```
mitogh: I we should change the order of this methods into: 

1. `renderImage`
1. `renderPlaceholder`
1. `renderLoading`

To make it top to bottom as function calls happens.

bordoni:
ready for another look

PR #88, opened 2018-06-12
Fix/106585 year month nav

paulmskim:
For ticket: https://central.tri.be/issues/106585

### DIFF
--- a/src/modules/elements/month/element.js
+++ b/src/modules/elements/month/element.js
@@ -107,6 +107,23 @@ export default class Month extends Component {
 
 	handleYearMonthChange = ( month ) => this.setState({ month });
 
+	getCaptionElement = ({ date, localeUtils }) => {
+		const { month } = this.state;
+
+		if ( date.getMonth() !== month.getMonth()) {
+			return this.renderCaption( date, localeUtils );
+		}
+
+		return (
+			<YearMonthForm
+				today={ today }
+				date={ date }
+				localeUtils={ localeUtils }
+				onChange={ this.handleYearMonthChange }
+			/>
+		);
+	}
+
 	renderCaption = ( date, localeUtils ) => (
 		<div className={'tribe-editor__daypicker-caption'} role="heading">
 			<div>
@@ -136,20 +153,7 @@ export default class Month extends Component {
 						before: today,
 					}
 				}
-				captionElement={ ({ date, localeUtils }) => {
-					if ( date.getMonth() !== month.getMonth()) {
-						return this.renderCaption( date, localeUtils );
-					}
-
-					return (
-						<YearMonthForm
-							today={ today }
-							date={ date }
-							localeUtils={ localeUtils }
-							onChange={ this.handleYearMonthChange }
-						/>
-					);
-				} }
+				captionElement={ this.getCaptionElement }
 			/>
 		);
 	}
--- a/src/modules/elements/month/element.js
+++ b/src/modules/elements/month/element.js
@@ -52,7 +52,6 @@ export default class Month extends Component {
 
 	constructor( props ) {
 		super( ...arguments );
-		this.handleYearMonthChange = this.handleYearMonthChange.bind(this);
 
 		this.state = {
 			...props,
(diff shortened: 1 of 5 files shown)


src/modules/elements/month/element.js
```diff
@@ -44,6 +52,7 @@ export default class Month extends Component {
 
 	constructor( props ) {
 		super( ...arguments );
+		this.handleYearMonthChange = this.handleYearMonthChange.bind(this);
```
mitogh: You can avoid this one by changing your method declaration from: 

```js 
handleYearMonthChange( month ) {
}
```

to 

```js
handleYearMonthChange = ( month ) => {
}
```

src/modules/elements/month/element.js
```diff
@@ -97,23 +106,52 @@ export default class Month extends Component {
...
 				selectedDays={ this.getSelectedDays() }
 				onDayClick={ this.selectDay }
+				onMonthChange={ this.handleYearMonthChange }
 				disabledDays={
 					{
-						before: new Date(),
+						before: today,
 					}
 				}
+				captionElement={ ({ date, localeUtils }) => {
```
mitogh: It might be worth to split this one into a method. I think the components inside of each conditional block as well.

Like: 

```js
if ( ... ) {
  return this.renderComponentA();
}

return this.renderComponentB();
```

src/modules/elements/month/element.js
```diff
@@ -97,23 +106,52 @@ export default class Month extends Component {
...
+					if ( date.getMonth() !== month.getMonth()) {
+						return (
+							<div className={'DayPicker-Caption'} role="heading">
```
mitogh: For class names we are using `tribe-editor` as a prefix for all elements or blocks that are part of the gutenberg editor. 

We are using BEM style as well example: 

```css
.tribe-editor__calendars--range {
}
```

```css
// Example above
.tribe-editor__{block}--{modifier} { 
}

// Full block
.tribe-editor__{block}__{element}--{modifier} {
}
```

src/modules/elements/year-month-form/element.js
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	const handleChange = ( e ) => {
+		const { year, month } = e.target.form;
+		onChange( new Date( year.value, month.value ) );
+	};
+
+	return (
+		<form className="DayPicker-Caption">
+			<select className="DayPicker-Caption__month" name="month" onChange={ handleChange } value={ date.getMonth() }>
```
mitogh: There's a lint rule that prevents lines to go over 100 chars length (if I'm not wrong); so it might be good if we can split this one into multiple lines: 

```js
<select
   className="..."
/>
```

src/modules/elements/month/element.js
paulmskim: I made these changes to the year month form. However, this part is an exact copy of the original table caption when we don't replace the caption element. Do we still want to change this and apply the styles again to the new selector?

mitogh:
I think for consistency it might be a good idea to follow the same pattern.

src/modules/elements/month/element.js
paulmskim: So this is the exact case I was referring to in our call earlier, where `this` is not referenced correctly due to this function being passed as a prop to another component. If I replace it as you have it above, it'll complain that `renderComponentA()` is not defined.

mitogh: Yeah you can define it as a variable like the example above:

```js
renderComponentA = () => {
}
```

src/modules/elements/month/element.js
mitogh: Yeah I think it make sense to have a consistency across this values.

src/modules/elements/month/element.js
```diff
@@ -97,23 +105,51 @@ export default class Month extends Component {
...
 				selectedDays={ this.getSelectedDays() }
 				onDayClick={ this.selectDay }
+				onMonthChange={ this.handleYearMonthChange }
 				disabledDays={
 					{
-						before: new Date(),
+						before: today,
 					}
 				}
+				captionElement={ ({ date, localeUtils }) => {
```
mitogh: It might be good idea to extract the function used in this property into is own method same as `onMonthChange`

PR #89, opened 2018-06-12
fix/107023 time separator all day

paulmskim:
For ticket: https://central.tri.be/issues/107023

### DIFF
--- a/src/modules/blocks/event-datetime/block.js
+++ b/src/modules/blocks/event-datetime/block.js
@@ -267,7 +267,7 @@ class EventDateTime extends Component {
 					<footer className="tribe-editor__subtitle__footer">
 						<section>
 							{ this.renderStartTimePicker() }
-							{ this.renderSeparator( 'time-range', 'tribe-editor__time-picker__separator' ) }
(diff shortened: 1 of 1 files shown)


src/modules/blocks/event-datetime/block.js
```diff
@@ -267,7 +267,7 @@ class EventDateTime extends Component {
 					<footer className="tribe-editor__subtitle__footer">
 						<section>
 							{ this.renderStartTimePicker() }
-							{ this.renderSeparator( 'time-range', 'tribe-editor__time-picker__separator' ) }
+							{ this.isAllDay() ? null : this.renderSeparator( 'time-range', 'tribe-editor__time-picker__separator' ) }
```
mitogh: I think this one is fine but wanted to flag that you can avoid the ternary operator by using: 

```js
{ this.isAllDay() && this.renderSeparator( 'time-range', 'tribe-editor__time-picker__separator' )  }
```

paulmskim: Oh yeah. I did this because it's part of the render, where `null` returns nothing.

PR #90, opened 2018-06-12
fix/107268 tec blocks on pages and posts

paulmskim:
For ticket: https://central.tri.be/issues/107268

### DIFF
--- a/src/Tribe/Editor.php
+++ b/src/Tribe/Editor.php
@@ -199,15 +199,26 @@ class Tribe__Events_Gutenberg__Editor {
 		return update_post_meta( $post, $this->key_flag_classic_editor, 1 );
 	}
 
+	/**
+	 * Check if post type is tribe events
+	 *
+	 * @since  0.2.2-alpha
+	 *
+	 * @return bool
+	 */
 	public function is_events_post_type() {
 		global $post, $typenow, $current_screen;
 
 		if ( $post && $post->post_type ) {
 			return $post->post_type === Tribe__Events__Main::POSTTYPE;
-		} elseif ( $typenow ) {
+		}
+
+		if ( $typenow ) {
 			return $typenow === Tribe__Events__Main::POSTTYPE;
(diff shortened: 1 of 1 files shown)


src/Tribe/Editor.php
```diff
@@ -199,6 +199,20 @@ public function flag_post_from_classic_editor() {
...
 
+	public function is_events_post_type() {
+		global $post, $typenow, $current_screen;
+
+		if ( $post && $post->post_type ) {
+			return $post->post_type === Tribe__Events__Main::POSTTYPE;
+		} elseif ( $typenow ) {
+			return $typenow === Tribe__Events__Main::POSTTYPE;
+		} elseif ( $current_screen && $current_screen->post_type ) {
+			$current_screen->post_type === Tribe__Events__Main::POSTTYPE;
```
mitogh: Missing a return here.

Also I think to avoid having `elseif` it might be better just to keep them as separate if statments as they just return early.

Like: 

```
if ( ... ) {
  return ...;
}

if( .... ) {
   return ...;
}
```

src/Tribe/Editor.php
```diff
@@ -199,6 +199,20 @@ public function flag_post_from_classic_editor() {
 		return update_post_meta( $post, $this->key_flag_classic_editor, 1 );
 	}
 
+	public function is_events_post_type() {
```
mitogh: Missing a dockblock here

src/Tribe/Editor.php
```diff
@@ -199,6 +199,31 @@ public function flag_post_from_classic_editor() {
...
+	 * Check if post type is tribe events
+	 *
+	 * @since  0.2.2-alpha
+	 *
+	 * @return bool
+	 */
+	public function is_events_post_type() {
+		global $post, $typenow, $current_screen;
+
+		if ( $post && $post->post_type ) {
```
bordoni: We should use the `Tribe__Admin__Helpers` methods.

PR #94, opened 2018-06-14
Fix/108306 dashboard closing issue

paulmskim:
@mitogh this is branched off your "closing dashboard when clicking on sidebar" branch, so review this once that branch is merged in.

On another note, I really dislike how WordPress handles its `Dropdown` component. If you look at the DOM, it is actually an absolutely positioned component **outside** of the edit post layout container and not a child or sibling of the actual button that opens and closes it. If possible, I'd be very much in favor of not using the `Dropdown` component.

For ticket: https://central.tri.be/issues/108306


mitogh:
OKay that makes sense; do you know any alternatives we can use instead?

Thanks @paulmskim

PR #95, opened 2018-06-14
fix/108305 tags event categories block

paulmskim:
For ticket: https://central.tri.be/issues/108305

### DIFF
--- a/src/modules/elements/terms-list/element.js
+++ b/src/modules/elements/terms-list/element.js
@@ -16,6 +16,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import './style.pcss';
 
 /**
  * Module Code
@@ -29,7 +30,7 @@ class TaxonomiesElement extends Component {
 		};
 	}
 
-	getTerms( parentId = null ) {
+	getTerms = ( parentId = null ) => {
 		if ( ! this.props.terms ) {
 			return [];
 		}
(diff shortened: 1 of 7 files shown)


src/modules/elements/terms-list/element.js
```diff
@@ -46,12 +47,15 @@ class TaxonomiesElement extends Component {
 		return terms.filter( term => term.parent === parentId );
 	}
 
-	getTermListClassName( level ) {
-		const { className } = this.props;
-		return `tribe-editor__terms__list tribe-terms__list-level-${ level } ${ className }`;
-	}
+	getTermListClassName = level => (
```
mitogh: Just for consistency reasons we should keep using the parenthesis around the param even if is only one.  `( level )`

src/modules/elements/terms-list/element.js
```diff
@@ -46,12 +47,15 @@ class TaxonomiesElement extends Component {
...
+		`tribe-editor__terms__list tribe-editor__terms__list--level-${ level }`
+	);
+
+	getTermListItemClassName = level => (
```
mitogh: Same as above.

src/modules/elements/terms-list/element.js
paulmskim: what linting rules are we using for ES6 here? I'm used to Airbnb's linting rules, and this is how they do it, but if we're doing something different, I'm down for that too.

mitogh: I think is a set of custom ones + WP Calypso.

- https://github.com/moderntribe/eslint-config-products/

paulmskim: I actually had a look at our eslint rules and, unless `products/react` has others specified, if we are using the default (`all`) for the `no-extra-parens` rule, the default for single parameter with `returnAssign` is as what is seen here: https://eslint.org/docs/rules/no-extra-parens

paulmskim: Sorry, disregard my last one. The `arrow-parens` rule on `always` (default) has parentheses around the parameters in all cases: https://eslint.org/docs/rules/arrow-parens

i'll make the changes.

PR #100, opened 2018-06-14
Fix/108304 website block fixes

paulmskim:
For ticket: https://central.tri.be/issues/108304


src/modules/blocks/event-website/block.js
```diff
@@ -86,26 +97,28 @@ class EventWebsite extends Component {
...
 		const { urlLabel, setAttributes } = this.props;
+		const text = urlLabel ? urlLabel : placeholder;
+		const textWidth = this.getTextWidth( text );
 
 		return (
-			<div key='tribe-events-website-label' className="tribe-editor__event-website">
-				<PlainText
-					id="tribe-events-website-link"
+			<div key='tribe-events-website-label' className="tribe-editor__event-website__label">
+				<input
```
mitogh: mm wonder why not use `RichText` here instead?

mitogh: Also mentioned this one as I've created a PR chainging the PlaintText to RichText

bordoni: PR mentioned by @mitogh https://github.com/moderntribe/events-gutenberg/pull/96

paulmskim:
alright @mitogh  back to you.

src/modules/blocks/event-website/block.js
```diff
@@ -2,25 +2,17 @@
  * External dependencies
  */
 import React from 'react';
+import ReactDOM from 'react-dom';
```
mitogh: Seems like this one is no longer used

src/modules/blocks/event-website/block.js
```diff
@@ -47,21 +41,23 @@ class EventWebsite extends Component {
 		super( ...arguments );
 	}
 
-	render() {
-		return [ this.renderButton(), this.renderControls() ];
+	onLabelChange = ( urlLabel ) => {
```
mitogh: It might be good to have the methods ordered in a way we can read top to bottom based on usage for example: 

```js
methodA() {
  methodB();
}

methodB() {
  methodC();
}

methodC() {
}
```

instead of: 

```js
methodC() {
}

methodB() {
  methodC();
}

methoA() {
  methodB();
}
```

paulmskim: I actually think it's typical to have it the way you described it second. Most components / blocks have the `render()` method last, and this method typically calls a number of other methods above it. If you look at how Gutenberg components do it, and other React approaches, they do it the same.

If you'd like to fix this, I'd like to do a pass on all the components and blocks we have, but I think we should set on a format and be consistent about it. I'd also be down to do this all at once rather than just this component and not the others.

mitogh: Okay wasn't aware of that just pointing as seems more logically (from my perspective).

I think we should talk a bit with @bordoni involved and some people from loxi to set an internal standard so we can update accordingly.

for now I think we can move forward.

paulmskim: No worries, for sure. Definitely let's set on a standard so we can clean up our components and blocks a bit.

PR #106, opened 2018-06-16
Apply feedback into the organizer and venue block

mitogh:


src/modules/blocks/event-organizer/details/style.pcss
```diff
@@ -1,88 +1,97 @@
...
+		padding: 5px 7px;
+		cursor: pointer;
+		border: none;
+
+		svg,
+		path {
+			fill: #8D949B;
+		}
+
+		&:hover {
```
paulmskim: i know this was copied over from what was already here, but if we're thinking of accessibility, a button should have a `hover`, `focus`, and `active` state.

src/modules/blocks/event-venue/block.js
```diff
@@ -210,6 +213,27 @@ class EventVenue extends Component {
...
 
+	maybeEdit = () => {
+		if ( this.hasVenue() && this.isDraft() ) {
+			return this.edit();
+		}
+	}
+
+	renderEditAction() {
+		const { isSelected } = this.props;
+		const { edit, create, loading, submit } = this.props;
```
paulmskim: you could combine the above 2 lines of destructuring into one.

src/modules/blocks/event-venue/block.js
```diff
@@ -260,12 +288,12 @@ class EventVenue extends Component {
 					<ToggleControl
 						label={ __( 'Show Google Maps Link' ) }
 						checked={ ! ! showMapLink }
-						onChange={ ( value ) => setAttributes( { showMapLink: value } ) }
+						onChange={ ( value ) => setAttributes( { showMapLink: !! value } ) }
```
paulmskim: Is there a need for the `!!` here? It looks like `onChange` provides a boolean as an input: https://github.com/WordPress/gutenberg/tree/master/components/toggle-control

src/modules/blocks/event-venue/block.js
```diff
@@ -260,12 +288,12 @@ class EventVenue extends Component {
...
 					/>
 					<ToggleControl
 						label={ __( 'Show Google Maps Embed' ) }
 						checked={ ! ! showMap }
-						onChange={ ( value ) => setAttributes( { showMap: value } ) }
+						onChange={ ( value ) => setAttributes( { showMap: !! value } ) }
```
paulmskim: same as above

src/modules/blocks/event-venue/style.pcss
```diff
@@ -65,24 +72,55 @@
...
+		background: transparent;
+		border: none;
+		cursor: pointer;
+
+		svg,
+		path {
+			fill: #8D949B;
+		}
+
+		&:hover {
```
paulmskim: same thing here with the `hover`, `focus`, and `active` states for button re: accessibility.

src/modules/blocks/event-venue/style.pcss
```diff
@@ -103,3 +141,25 @@
...
```
paulmskim: same as above on the `hover`, `focus`, and `active` states.

src/modules/elements/search-or-create/style.pcss
```diff
@@ -19,19 +19,23 @@
...
+		fill: #E2E4E7;
+		width: 22px;
+	}
 
+	path {
+		fill: #E2E4E7;
+	}
 }
 
 .tribe-editor__soc__input__container--active {
```
paulmskim: since you've been fixing spacing, maybe you can fix the indentation from here to line 80 as well.

src/modules/blocks/event-organizer/details/style.pcss
mitogh: Great point, agree with that however I don't think we have styles or something for those states. 

We already [have a ticket](https://central.tri.be/issues/108534) to handle some accessibility feedback

Maybe we can revisit all the `buttons` or the clickable items to handle this scenarios as well.

I think we should revisit other items as well and ask for a style guide for those states.

Thanks

src/modules/blocks/event-venue/style.pcss
mitogh: Replied above.

src/modules/blocks/event-venue/style.pcss
mitogh: ⬆️

src/modules/blocks/event-venue/block.js
mitogh: It was a string before due a PHP error but forgot to update it as is always a boolean now.

Thanks

src/modules/blocks/event-venue/block.js
mitogh: 💯

src/modules/blocks/event-organizer/details/style.pcss
paulmskim: Sounds good

src/modules/elements/search-or-create/style.pcss
paulmskim: Before you merge, if you could just fix the spacing from this comment, that'd be awesome.

mitogh: This one has been fixed I think is not reflected on the diff.

PR #108, opened 2018-06-18
Improve layout styles for block to avoid hide controls

mitogh:


src/modules/editor/style.pcss
```diff
@@ -3,20 +3,52 @@
  */
 .gutenberg-editor-page.post-type-tribe_events {
 	.edit-post-visual-editor {
+		.editor-post-title {
+			@media screen and (max-width: 782px) {
```
paulmskim: instead of using `max-width` here, set the margin as `margin: 0 auto` without any media queries, then define something for `min-width: 783px`.

src/modules/editor/style.pcss
```diff
@@ -3,20 +3,52 @@
...
+
 		/**
 		 * On our elements we want 900px instead of the default
 		 * possibly keep an eye to how to do it on PHP
 		 */
 		.editor-block-list__block,
 		.editor-default-block-appender,
 		.editor-post-title {
-			max-width: 900px;
+			@media screen and (max-width: 782px) {
```
paulmskim: same thing here. the way this actually works is that on mobile up to 782px, max width is `calc( 100% - 60px ), but from 783px to 1300px, there is no max width (not sure if this is what you intended, maybe it is). Either way, we should be working with mobile styles as default then working up.

mitogh:
Changes has been applied.

mitogh:
Another approval is required after the conflicts resolutions.

PR #109, opened 2018-06-18
Fix/108535 delete block disconnect data

paulmskim:
For ticket: https://central.tri.be/issues/108535


mitogh:
Conflicts here @paulmskim

PR #115, opened 2018-06-20
Fix styles to follow styleguide

mitogh:
Ticket: https://central.tri.be/issues/


src/modules/blocks/event-datetime/style.pcss
```diff
@@ -24,7 +24,7 @@
 
 .tribe-editor__subtitle {
 
-	.tribe-editor__timepicker {
+	.tribe_editor__timepicker {
```
paulmskim: is this intentional? if so, can we change the timepicker class instead, to keep with BEM style?

mitogh: Amm I think is BEM already or do you mean `tribe_editor__time__picker`?

paulmskim: ah i mean everywhere else, it's `tribe-editor__` rather than `tribe_editor__`

mitogh: Seems like the diff is not updating this one correctly.

However this one has been updated:

- https://github.com/moderntribe/events-gutenberg/blob/fix/108762-css-states/src/modules/blocks/event-datetime/style.pcss#L27

A note on this one is we are using *tribe-editor* instead of *tribe_editor* as prefix for editor styles, as if we use *tribe_editor* it should be *tribe__editor* but the whole *tribe-editor* is used as the prefix instead.

Let me know if you have any questions.

mitogh:
Changes has been applied

src/modules/blocks/event-datetime/style.pcss
paulmskim: @mitogh the commit below doesn't update this file, which is why the diff isn't updating for this one.

mitogh: Woops that's right! Thanks for flagging this one.

mitogh:
Changes has been applied thanks @paulmskim

PR #116, opened 2018-06-21
Update the structure of the store into own action

mitogh:
Ticket: https://central.tri.be/issues/108803


src/modules/data/organizers/index.js
```diff
@@ -1,163 +1,23 @@
...
-			}
 
-			const body = await apiRequest( { path: `/wp/v2/${ POST_TYPE }/${ organizer }` } );
-			dispatch( STORE_NAME ).setPost( id, body );
-		},
-	},
-};
-export const store = registerStore( STORE_NAME, details );
+export const store = registerStore( STORE_NAME, {
+	reducer: combineReducers( reducers ),
```
paulmskim: if we only have one reducer, we don't need the `combineReducers()` here.

src/modules/data/organizers/resolvers.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Wordpress dependencies
+ */
+import { dispatch } from '@wordpress/data';
+const { apiRequest } = wp;
+
+import { POST_TYPE, STORE_NAME } from './index';
+
+export async function getDetails( state, id, organizer ) {
```
paulmskim: same thing here on arrow functions: `export const getDetails = async ( state, id, organizer ) => { ... }`

src/modules/data/organizers/selectors.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * External dependencies
+ */
+import { get, pick, map, isEmpty } from 'lodash';
+
+export function getPosts( state, id ) {
```
paulmskim: same thing with es6 arrow functions here.

src/modules/data/details/index.js
```diff
@@ -39,7 +40,7 @@ export const DEFAULT_STATE = {
 
 export const STORE_NAME = 'tec.details';
 
-const details = {
+export const store = registerStore( STORE_NAME, {
 	reducer( state = DEFAULT_STATE, action ) {
```
paulmskim: could we also move the reducer to a separate file? I think for the organizers you've done it that way.

src/modules/data/search/organizer/selectors.js
```diff
@@ -0,0 +1,7 @@
+export function fetchDetails( state, organizers ) {
```
paulmskim: probably no need to pass in organizers here? also we can use shorthand arrow function notation here.

src/modules/data/venue/resolvers.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+import { POST_TYPE, STORE_NAME } from './index';
+
+export async function getDetails( state, id ) {
+	if ( state.id === id ) {
+		return state;
+	}
+
+	apiRequest( { path: `/wp/v2/${ POST_TYPE }/${ id }` } )
+		.then( ( body ) => {
```
paulmskim: we could just return the dispatch function here

src/modules/data/venue/resolvers.js
```diff
@@ -0,0 +1,26 @@
...
+ * Internal dependencies
+ */
+import { POST_TYPE, STORE_NAME } from './index';
+
+export async function getDetails( state, id ) {
+	if ( state.id === id ) {
+		return state;
+	}
+
+	apiRequest( { path: `/wp/v2/${ POST_TYPE }/${ id }` } )
```
paulmskim: a thought here, it doesn't have to happen in this PR but I think we can have a constant for `WP_API_PREFIX = 'wp/v2/'` or something like that so that if wp's api does change and goes to v3 or something completely different, we can change it easily. also it removes error from typing and such.

src/modules/data/organizers/index.js
```diff
@@ -1,163 +1,23 @@
...
-				return post;
-			}
 
-			const body = await apiRequest( { path: `/wp/v2/${ POST_TYPE }/${ organizer }` } );
-			dispatch( STORE_NAME ).setPost( id, body );
-		},
-	},
-};
-export const store = registerStore( STORE_NAME, details );
+export const store = registerStore( STORE_NAME, {
```
paulmskim: would love to discuss this with you, but I'm curious why we need more than 1 store in this application. frankly, the only things `registerStore()` needs are the store name and reducers as everything should be described through the reducers. I'm not seeing the benefit to having multiple stores, but if there's a good reason then I'm OK with this.

src/modules/data/search/organizer/selectors.js
mitogh: Is not required however in order to keep the signature for resolvers is requires as the selector is passed into the resolver that uses the param.

src/modules/data/organizers/index.js
mitogh: Multiple reasons basically.

1. In order to keep store outside of any pollution not related to the data for instance: Event details, Searches, Multiple Blocks States

2. For searches we might be able to get away from that store by using HOC so we can plan to refactor that behavior at some point for now we need a place to hold the state of searches / drafts and having that into the main Event store just doesn't fit the needs as mentioned before we can get ride of that with HOC so when that happen we don't have to touch the main store.

3. Some blocks rely on other blocks for his state for instance the organizers where organizers blocks and event details are related having 1 and 2 into the same place of logic will prevent us for moving forward with any type of refactor there.

Ideally we just need 1 store but in order for us to iterate quickly we set up a couple as mechanism for hold data in the fly.

src/modules/data/venue/resolvers.js
mitogh: That's a good idea.

src/modules/data/details/actions.js
```diff
@@ -0,0 +1,133 @@
+export function addOrganizer( organizer ) {
```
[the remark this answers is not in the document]
mitogh: TLDR: Basically for better readability and consistency.

1. Arrow functions are great mechanism to improve readability, and in most cases are helpful however when your function became too long in order to fit it in a single line doesn't feel right to me as we should favor readability because having an arrow function and a regular function does not have any difference other than just less verbosity and if the file is minified and there's no real value I think we should favor readability for consistency.

2. Consistency, I think if we follow this patter of using arrow functions for most of the actions we should apply the same pattern for all actions otherwise things get confusing easier as when we decided when to apply each pattern. And in some cases we don't have actions that does not fit in a single line so we might end up with multiple lines and at that point using the arrow function does not provide much difference as using a regular function so I think in order to avoid having in a scenario when we have to decide when is a good time to use an arrow function might be better to use regular functions.

src/modules/data/organizers/resolvers.js
mitogh: Same as above.

src/modules/data/organizers/selectors.js
mitogh: Same as above.

src/modules/data/details/index.js
mitogh: Right!

src/modules/data/details/actions.js
```diff
@@ -0,0 +1,133 @@
...
+	return {
+		type: 'ADD_ORGANIZER',
```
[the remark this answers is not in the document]
mitogh: In this one I see there two places where the action might be used but I don't think see the benefit of using a constant here if the the store and reducer, actions and reducers are the only ones interacting with this value. 

Plus I haven't see a pattern where the type `action.type` is been set into a constant. 

As I think creates and unnecessary overhead.

src/modules/data/organizers/index.js
mitogh: Yeah, however as the reducer is multidimensional object having `combineReducers` makes it more clear, otherwise the default state of the store is: 

```js
blocks: {
 ....defaultState
}
```

src/modules/data/details/actions.js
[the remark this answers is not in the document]
paulmskim: Sure, I totally get it that we shouldn't fit it into one line. What I meant to write was:
```
export const addOrganizer = ( organizer ) => ( {
    type: 'ADD_ORGANIZER',
    organizer,
} );
```
My main point here is that we should pick one format and use it. I don't know if this should be in a different ticket or not, but if we're using different formats, it'll become confusing for anyone else that may come into the project to work on it.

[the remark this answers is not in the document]
mitogh: Yeah I think that's something we need to address as well as is not defined correctly when and where we should use which.

PR #124, opened 2018-06-29
Fix/108301 price block issues

paulmskim:
For ticket: https://central.tri.be/issues/108301

PR #125, opened 2018-06-29
Fix/108300 date time block fixes

paulmskim:
For ticket: https://central.tri.be/issues/108300


mitogh:
On hold due: 

- https://github.com/moderntribe/events-gutenberg/pull/124#issuecomment-401234358

paulmskim:
@mitogh could you review this when you get a chance? thanks!

mitogh:
Seems like you have a conflict to solve on `readme.md` @paulmskim

src/modules/blocks/event-datetime/block.js
```diff
@@ -202,6 +219,11 @@ class EventDateTime extends Component {
 			.isSame( toMoment( end || this.props.end ), 'day' );
 	}
 
+	isSameYear( start, end ) {
```
mitogh: Seems like you can simplify the conditional fallback on the function by using default params

```js
isSameYear( start = this.props.start, end = this.props.end ) {
 ...
}
```

src/modules/editor/utils/date.js
```diff
@@ -12,13 +12,16 @@ export const FORMATS = {
 	TIME: 'HH:mm:ss',
 	DATE_TIME: 'YYYY-MM-DD HH:mm:ss',
 	WP: {
+		...formats,
 		time: 'g:i a',
-		date: 'F j',
+		date: 'F j, Y',
 		datetime: 'F j, Y g:i a',
-		...formats,
+		dateNoYear: 'F j',
```
mitogh: We need to overwrite the default with `formats` which are the values defined on the WP editor.

src/modules/blocks/event-datetime/block.js
paulmskim: this was basically copied from your `isSameDay()` function, but checking year. I will make this change for both.

paulmskim:
back to you @mitogh

mitogh:
👍 @paulmskim

PR #129, opened 2018-07-02
Update classic block to connect with the new store

mitogh:
- Remove `withSelect` & `withDispatch` from classic block
- Update block to connect with then new store


mitogh:
Requires #126

PR #130, opened 2018-07-03
Add "Event Links" block into the store

mitogh:
In order to have sync copies of blocks we need to register all blocks into the store

**This PR requires:**  https://github.com/moderntribe/events-gutenberg/pull/129


src/modules/blocks/classic-event-details/block.js
```diff
@@ -6,6 +6,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { connect } from 'react-redux';
 import { compose, bindActionCreators } from 'redux';
+import AutosizeInput from 'react-input-autosize';
```
bordoni: I am so happy to see this.

PR #131, opened 2018-07-03
Compatibility for Event Tickets

bordoni:
_Ref:_ [C#105158](http://central.tri.be/issues/105158)
---

### DIFF
--- a/src/Tribe/Compatibility/Tickets.php
+++ b/src/Tribe/Compatibility/Tickets.php
@@ -47,10 +47,10 @@ class Tribe__Events_Gutenberg__Compatibility__Tickets {
 		do_action( $hook );
 		$form = ob_get_clean();
 
-		if ( false !== strpos( $hook, 'before' ) ) {
-			return $form . $content;
-		} else {
+		if ( false === strpos( $hook, 'before' ) ) {
 			return $content . $form;
+		} else {
+			return $form . $content;
 		}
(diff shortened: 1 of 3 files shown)


src/Tribe/Compatibility/Tickets.php
```diff
@@ -0,0 +1,56 @@
...
+			return $content;
+		}
+
+		$hook = tribe( 'tickets.rsvp' )->get_ticket_form_hook();
+
+		ob_start();
+		do_action( $hook );
+		$form = ob_get_clean();
+
+		if ( false !== strpos( $hook, 'before' ) ) {
```
mitogh: I think by inverting the conditional here will make it more explicit to read for example: 

```js
if ( false === strpos( $hook, 'before' ) ) {
	return $content . $form;
} else {
	return $form . $content;
}
```

src/Tribe/Provider.php
```diff
@@ -47,12 +49,32 @@ public function register() {
...
 		 * @todo  Remove this later on
 		 */
 		tribe( 'gutenberg.editor' )->assets();
 	}
 
+	/**
+	 * Initializes the correct classes for when Tickets is active.
+	 *
+	 * @since  0.2.4-alpha
+	 *
```
mitogh: Missing `@return`

PR #132, opened 2018-07-03
Allow sync of blocks

mitogh:
Prevent new copies of blocks remove the state of created blocks already on the editor.

### DIFF
--- a/src/modules/editor/hoc/with-save-data.js
+++ b/src/modules/editor/hoc/with-save-data.js
@@ -13,6 +13,11 @@ import {
 import isShallowEqual from '@wordpress/is-shallow-equal';
 
 /**
+ * Wordpress dependencies
+ */
+import { select } from '@wordpress/data';
+
+/**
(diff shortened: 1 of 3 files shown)


src/modules/editor/hoc/with-save-data.js
```diff
@@ -71,6 +88,15 @@ export default ( selectedAtrributes = null ) => ( WrappedComponent ) => {
 				} );
 			}
 
+			countSameBlocks( ) {
+				const { name = '' } = this.props;
+				const blocks = select( 'core/editor' ).getBlocks()
+					.map( ( block ) => block.name )
+					.filter( ( blockName ) => blockName === name );
```
paulmskim: rather than doing `map()` then `filter()` here, I would suggest, instead, to do `filter( ( block ) => block.name === name );` so you don't run through the array twice.

mitogh:
Changes has been applied, let me know if you have any other concerns @paulmskim

PR #137, opened 2018-07-05
Implement organizer block view

juanfra:
🎫 https://central.tri.be/issues/109047


.gitignore
```diff
@@ -2,6 +2,7 @@
 build
 src/resources/*
 !src/resources/icons
+!src/resources/css/views.css
```
bordoni: We don't want to push compiled files into the repo. The person needs to run `npm run build` to get it going.

juanfra: This CSS file contains the necessary styling for the FE views. The idea is that they could have it out of the box, so the event they create look on the front end as what they have in the Gutenberg editor.

juanfra: We have to include one file for that, in one way or the other. Else, we cannot style the blocks on the front-end.

juanfra: (unless we want to make this extension purely accesible and usable by developers who will clone the repo and then build it themselves)

bordoni: We build the zips for users just like we do with our plugins, if you are cloning you need to run the building process

juanfra: perfect, thanks! Didn't know about the zip building, and as in the rest of the plugins the public stylesheets are tracked, thought we had to. 👍

src/modules/views/event-organizer/style.pcss
```diff
@@ -3,3 +3,25 @@
  *
  * @since  TBD
  */
+.tribe-block__organizer__details {
+	margin: 15px 0;
+	position: relative;
+
+	h3 {
```
paulmskim: could we target this and the `p` below with classes rather than the tags? just as a general practice, it makes it easier to cascade when necessary.

juanfra: Great, done!

juanfra:
Merge conflict with readme, I need a new review, please.

PR #139, opened 2018-07-06
Implement the event website block frontend view

juanfra:
🎫 https://central.tri.be/issues/109051

Implement the event website block frontend view


src/modules/views/event-website/style.pcss
```diff
@@ -3,3 +3,24 @@
  *
  * @since  TBD
  */
+
+.tribe-block__event-website {
+	padding: 4% 20px;
+	align-items: center;
+	border: 1px solid #E1E3E6;
+	display: inline-flex;
+	font-family: 'Helvetica', 'sans-serif';
```
bordoni: We should avoid font-family for front-end stuff.

src/modules/views/event-website/style.pcss
```diff
@@ -3,3 +3,24 @@
...
+	justify-content: center;
+	min-height: 54px;
+	padding: 0 20px;
+	font-size: 1.125rem;
+	background-color: #009FD4;
+	max-width: 100%;
+	color: #fff;
+
+	&,
```
bordoni: Typo?

bordoni:
Also changelog

src/modules/views/event-website/style.pcss
juanfra: 👍

src/modules/views/event-website/style.pcss
juanfra: nope, actually a code that I left before adding the color to the class. `&` means self, so the idea was to assign the color property to both the class and the hover state. Removing that.

juanfra:
> Should we split that Border, padding and Background color into something re-usable?

Like a "size" class, and border class or something like that?
i.e
`.our-button.medium {
   padding: X Y;
}
.our-button.bordered {
   border: 1px solid black;
}
`

bordoni:
Possibly into a global rule for that style of block. Not sure tho.

src/views/blocks/event-website.php
```diff
@@ -1,21 +1,24 @@
...
-    class="tribe-block__website"
-    href="<?php echo esc_url( $attributes['href'] ); ?>"
-    target="<?php echo esc_attr( $target ); ?>"
-    <?php if ( '_blank' === $target  ) : ?> rel="noopener noreferrer" <?php endif; ?>
-    >
-    <?php echo esc_html( $attributes['urlLabel'] ); ?>
-</a>
+<div class="tribe-block tribe-block__event-website">
+    <a
+        class="blue bordered"
```
bordoni: We need to avoid these very generic classes since this is on the front-end.

juanfra: I thought that's what you asked.

PR #142, opened 2018-07-06
Implement attributes wrapper for views

juanfra:
🎫 https://central.tri.be/issues/110022

@bordoni this is a WIP, not sure if I got you 100% on this one. Perhaps you can check it and we can discuss any further in a HO. Pushing this just for you to check the changes.


src/Tribe/Template.php
```diff
@@ -33,5 +33,20 @@ public function attributes( $default_attributes = array() ) {
...
+
+	/**
+	 * Return a specific attribute
+	 *
+	 * @since TBD
+	 *
+	 * @param array $default_attributes
+	 * @return array
+	 */
+	public function attr( $index, $default_attributes = array() ) {
```
mitogh: Wonder if this just return a single attribute value we should specify the default value for that attribute only not the whole block

src/Tribe/Blocks/Abstract.php
```diff
@@ -23,6 +23,19 @@ public function name() {
...
+	 * Return the block attributes
+	 *
+	 * @since  TBD
+	 *
+	 * @param  array $attributes
+	 *
+	 * @return array
+	*/
+	public function attributes( $params = array() ) {
+		return apply_filters( 'tribe_block_attributes_defaults_' . $this->slug(), $params );
```
bordoni: Add docblock, also pass the `$this` variable. As the last param.
Lets name this one without the Slug, so users can target all blocks via one filter if needed.

src/Tribe/Template.php
bordoni: Agree with this.

src/Tribe/Blocks/Abstract.php
bordoni: Both filters can be applied here, just make sure we create an abstract method for the `default_attributes` so we can setup em on each block.

juanfra:
Thanks @mitogh and @bordoni! I just pushed some changes. Not totally sure if that's what you wanted @bordoni, but I think this gives us the flexibility to set the defaults per block and to make them hookable.

mitogh:
Maybe we can remove this branch @juanfra ?

juanfra:
Absolutely, sorry for that @mitogh 🎉

mitogh:
🏆 @juanfra

PR #143, opened 2018-07-07
Refactor stores

mitogh:
Remove dependency from the Gutenberg wrapper for stores and use redux instead


paulmskim:
I'm not sure where to get started with this, it's just way too much code to go through. @bordoni thoughts? Should we just merge and tackle problems from there?

bordoni:
I am taking a look to check for anything for us to talk in the future. Would love you could do the same here @paulmskim, create the tickets and link em here, if you end up doing it.

mitogh:
It would be great if we collect all of those in a document and discuss in call to see which ones are worth to keep them as ticket. 

Cc @bordoni @paulmskim

PR #144, opened 2018-07-07
fix/108550 venue block fe

paulmskim:


src/views/blocks/parts/map.php
```diff
@@ -1,4 +1,9 @@
 <?php
+
+if ( ! tribe_embed_google_map() ) {
```
bordoni: We should save it in a variable and than check it since we most likely will use it later no?

paulmskim: this is the only place it's used. if embed google map = true, then continue with view markup, otherwise, bail early.

PR #145, opened 2018-07-09
Fix organizer block on initialize and venue defaults

mitogh:
- Set default values for Location block
- Fix Organizer when created with empty value


mitogh:
Sorry I might need another set of eyes; thank you

PR #148, opened 2018-07-10
Fix count of blocks of the same type

mitogh:
Prevent the initialization of blocks that are of the same type and are not isolated, which means they should be initialized only once per block


mitogh:
Merge from `master` requires another look.

Thanks

PR #156, opened 2018-07-12
Fix/108306 date time dashboard closing

paulmskim:
Move logic out of dashboard element and into blocks to pass logic down.
Fix dashboard closing issue when clicking on block and settings.

@mitogh something in here prevents the cost description from saving. could you have a look through at why this is happening?


src/modules/blocks/event-price/block.js
```diff
@@ -194,15 +195,41 @@ class EventPrice extends Component {
...
 
-	onCloseDashboard = () => {
-		this.setState( ( state ) => {
-			const { open } = state;
-			if ( ! open ) {
-				return null;
+	/* TODO: This needs to move to logic component wrapper */
+	onKeyDown = ( e ) => {
+		const ESCAPE_KEY = 27;
+		if ( e.keyCode === ESCAPE_KEY ) {
```
bordoni: Can we invert this?

src/modules/blocks/event-datetime/block.js
```diff
@@ -83,21 +84,27 @@ class EventDateTime extends Component {
...
 	};
 
-	constructor() {
-		super( ...arguments );
+	componentDidMount() {
+		document.addEventListener( 'keydown', this.onKeyDown );
+		document.addEventListener( 'click', this.onClick );
+	}
+
+	componentDidUnmount() {
```
nealfennimore: This should be `componentWillUnmount` unless there's something I'm missing with this build

src/modules/blocks/event-price/block.js
```diff
@@ -32,21 +32,30 @@ import {
...
-		super( ...arguments );
+	componentDidMount() {
+		document.addEventListener( 'keydown', this.onKeyDown );
+		document.addEventListener( 'click', this.onClick );
+	}
 
-		this.state = {
-			open: false,
-		};
+	componentDidUnmount() {
```
nealfennimore: `componentWillUnmount` here too :)

src/modules/blocks/event-datetime/block.js
paulmskim: you're right, thanks.

mitogh:
Wonder why we need to move that logic the components instead @paulmskim ?

Is this because every component can have other components that might not trigger the out, if so you `Dashboard` has a property to extend this.

src/modules/elements/dashboard/element.js
```diff
@@ -36,203 +35,32 @@ export default class Dashboard extends Component {
...
-		return searchParent( node, testNode => testNode.classList.contains( 'edit-post-sidebar' ) );
-	}
-
-	/**
-	 * Validate if the clicked node is part of any of the targets provided by the component.
-	 *
-	 * @param {DOMNode} node - The clicked node
-	 * @returns {boolean} true if the node is any of the targets
-	 */
-	validateChildren( node ) {
```
mitogh: It seems like you only need to provide a value to targets here to avoid duplication.

paulmskim:
@mitogh we want to move as much of the logic out of presentational components and smaller elements as possible, and move it to a higher level. We have different logic that depends on which block the component is used in and in doing this, it makes the smaller level components as clueless to the logic as possible, meaning it only needs props to be passed in to do what it needs to do.

mitogh:
Right! 

However we are doing the opposite here by attaching the behavior into both components. 

And seems like the property targets should be enough to update this to avoid having the same behavior in both components.

I'm okay with moving into a HOC instead to make it more reusable.

paulmskim:
As we want to make the application more maintainable and scalable, we need to make a clean separation between presentation and logic.

I'd be happy to chat about this with you and @nealfennimore but as far as I know, this is the typical pattern used when building in React to use components within components and separate concern between presentation and logic.

mitogh:
Absolutely and I think we should use HOC as much as possible, however I don't see this pattern applied on this PR.

I think either: 

1. We build the HOC and apply on this PR
2. Or we just use the `targets` property for now and apply the HOC later.

mitogh:
A FYI we already have a few of them over here: 

- https://github.com/moderntribe/events-gutenberg/tree/master/src/modules/editor/hoc

paulmskim:
So there are a couple main reasons why I did not use HOCs here:

1. We're calling `Dashboard` only in 2 blocks. I think that we would be prematurely optimizing if we were trying to keep things DRY just because we're using `Dashboard` twice. If we used it in 4 or 5 blocks, I agree we can think about abstracting out the logic.
2. The checks for `isTargetIn____` are not the same within the same block and across blocks. As each blocks have their own requirements for checking if the click target is within the active area, trying to abstract this out feels like we'd be forcing it.

Also, my main reason for not using target here is because that's putting the logic back into the element, rather than keeping it stateless and logic-less. Instead, if we just pass it an `open` prop, it knows to either open or close, nothing more.

Again, I would really like to discuss the reasoning behind these decisions here, rather than messaging back and forth. I'd also like to loop in Neal as I hear he's the React dev in products and I'm sure he'd have some good insight into this.

src/modules/elements/dashboard/element.js
```diff
@@ -30,220 +28,37 @@ export const directions = {
...
-
+const Dashboard = ({ className, direction, open, children }) => {
+	const getContainerClass = () => {
 		return classNames(
 			'tribe-editor__dashboard__container',
 			`tribe-editor__dashboard__container--${ direction }`,
-			{ 'tribe-editor__dashboard__container--overflow': overflow },
-			{ 'tribe-editor__dashboard__container--open': this.props.open },
+			{ 'tribe-editor__dashboard__container--open': open },
 			...className,
```
nealfennimore: I know this is an older piece of code here, but I'm trying to understand why this is using the spread operator if `className` is supposed to be string :question:

mitogh:
_After recover my eyes from the colors of the example 🤣_

> I have question concerning the target based methods. It looks like those target methods are only for determining whether or not we should close the block itself, is that correct?

Yes that's correct and I think following the HOC would be better to remove logic from the component on this one.

do you think we could implement something like this: codepen.io/nealfennimore/pen/WKQaKy instead? It would remove the target methods entirely as well as it's interdependency with CSS classes.

I think that would be great problem if the possible targets were elements with "clickable" items but in most cases _sidebar_ contains items that require click to work with no side effects or without stoping propagation.

I think we should take this PR as an opportunity to introduce the HOC mentioned on: https://github.com/moderntribe/events-gutenberg/pull/51#discussion_r181905715 from my POV I think it might be the best approach here.

nealfennimore:
@mitogh no doubt I wanted to get a prototype up fast - I probably should have put :dark_sunglasses: disclaimer in there :)

Okay, I wasn't entirely sure on the sidebar there. I'm guessing that's outside our control and thus we would need clicks to propagate in some cases. So, we'll probably have to do the HOC approach.

Looks like we can control which elements outside the HOC will prevent closing the component when clicking outside of it: https://github.com/Pomax/react-onclickoutside#marking-elements-as-skip-over-this-one-during-the-event-loop. I just hope that acts to prevent clicks from closing the component _within_ whatever we ignore as well.

nealfennimore:
Sounds good :)

PR #161, opened 2018-07-13
Add in @babel/runtime package to fix build error

nealfennimore:
Had trouble building assets without this. Maybe the newest `@wordpress/hooks` is dependent on babel 7? 

```
ERROR in ./node_modules/@wordpress/hooks/build-module/createHooks.js
Module not found: Error: Can't resolve '@babel/runtime/core-js/object/create' in '/home/neal/Projects/ModernTribe/events-gutenberg/node_modules/@wordpress/hooks/build-module'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
 		"validator": "^9.4.1"
 	},
 	"devDependencies": {
+		"@babel/runtime": "^7.0.0-beta.53",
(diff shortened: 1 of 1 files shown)


mitogh:
Weird.

Haven't faced that problem, but seems like a good call

PR #162, opened 2018-07-16
Add tests for utils functions

mitogh:
Add new tests for the utils functions

🎁 ✅

### DIFF
--- a/jest.globals.js
+++ b/jest.globals.js
@@ -1,2 +1,1 @@
 global.jQuery = () => {};
-global.window = {};
--- a/jest.globals.js
+++ b/jest.globals.js
@@ -1,2 +1,2 @@
-global.jQuery = () => {
-}
+global.jQuery = () => {};
+global.window = {};
--- a/src/modules/editor/settings.js
+++ b/src/modules/editor/settings.js
@@ -4,10 +4,6 @@
 import { get, has } from 'lodash';
 
 /**
- * WordPress dependencies
- */
-
-/**
  * Internal dependencies
  */
 import { settings } from 'utils/globals';
--- a/src/modules/editor/settings.js
+++ b/src/modules/editor/settings.js
@@ -6,7 +6,9 @@ import { get, has } from 'lodash';
 /**
  * Internal dependencies
  */
-import { settings } from 'utils/globals';
+import { settings as globalSettings } from 'utils/globals';
+
+let settings = {};
 
 /**
  * Module Code
@@ -14,13 +16,21 @@ import { settings } from 'utils/globals';
  * @return {object} An object with the global settings
  */
 export function getSettings() {
-	return settings;
+	return globalSettings || {};
 }
 
 export function getSetting( key, defaultValue = undefined ) {
-	return get( getSettings(), key, defaultValue );
+	return get( settings.getSettings(), key, defaultValue );
 }
 
 export function hasSetting( key ) {
-	return has( getSettings(), key );
+	return has( settings.getSettings(), key );
 }
+
+settings = {
+	getSettings,
+	getSetting,
+	hasSetting,
+};
+
+export default settings;
--- a/src/modules/editor/settings.js
+++ b/src/modules/editor/settings.js
@@ -4,33 +4,18 @@
 import { get, has } from 'lodash';
 
 /**
- * Internal dependencies
- */
-import { settings as globalSettings } from 'utils/globals';
-
-let settings = {};
-
-/**
  * Module Code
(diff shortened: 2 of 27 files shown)


src/modules/editor/settings.js
```diff
@@ -3,28 +3,34 @@
...
-	return get( getSettings(), key, defaultValue );
+	return get( settings.getSettings(), key, defaultValue );
 }
 
 export function hasSetting( key ) {
-	return has( getSettings(), key );
+	return has( settings.getSettings(), key );
 }
+
+settings = {
```
nealfennimore: Why not just:

nealfennimore: Why not just do:

```js
export default {
	getSettings,
	getSetting,
	hasSetting
}
```

src/modules/editor/tests/settings.test.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+import settings, { getSettings, getSetting, hasSetting } from '../settings';
+
+describe( 'Tests for settings.js', () => {
+	test( 'It should return an object as the settings', () => {
+		expect( getSettings() ).toEqual( {} );
+	} );
+
+	test( 'It should return the default', () => {
+		const spy = jest.spyOn( settings, 'getSettings' );
```
nealfennimore: We should really be testing against global (`window`) here, rather than getting a spy to return our wanted output.

src/modules/editor/utils/globals.js
```diff
@@ -1,7 +1,14 @@
+/**
+ * Internal dependencies
+ */
 import { get } from 'lodash';
 
-const global = window || {};
+export const global = window || {};
```
nealfennimore: Is there an environment where window would not be defined?

src/modules/editor/settings.js
mitogh: In order to bypass the rule about having a variable defined before the usage of the variable as we are using `variable.getSettings` inside of the methods. So we can declare it at the top and export it at the bottom with the right values.

src/modules/editor/tests/settings.test.js
mitogh: In this particular case those variable used on window are injected via PHP, do you think that one might be a use case for this as well?

jest.globals.js
```diff
@@ -1,2 +1,2 @@
-global.jQuery = () => {
-}
+global.jQuery = () => {};
+global.window = {};
```
nealfennimore: `global` should already be a mocked `window`

src/modules/editor/tests/settings.test.js
nealfennimore: Yes, definitely! Mutate the window object to include our needed properties, and then reset it when we're done

src/modules/editor/settings.js
```diff
@@ -3,28 +3,34 @@
...
  * Module Code
  *
  * @return {object} An object with the global settings
  */
 export function getSettings() {
-	return settings;
+	return globalSettings || {};
 }
 
 export function getSetting( key, defaultValue = undefined ) {
```
nealfennimore: No need to specify a default of `undefined` here. If there's no value the default will always be `undefined`

src/modules/editor/settings.js
nealfennimore: Ah I see now - this should work the same way. How about just doing this in the get/has functions?

```js
return get( getSettings(), key, defaultValue );
```

src/modules/editor/utils/globals.js
mitogh: I don't think so but wasn't aware this was defined on test environment, also wanted to make sure we are able to overwrite in case it might be required.

src/modules/editor/settings.js
mitogh: Yeah that was the initial signature of this function, however in order to mock `getSettings` value or `spyon` we need to be able to replace `getSettings` during tests execution so I added a default object with the properties which is replaced during test execution so `getSettings` can have a mock value.

nealfennimore: Do you think you could implement something like this: https://github.com/moderntribe/calendar-saas-app/blob/027eae7f98851df816440298cfec083faa371d66/src/client/scripts/EventCreate/containers/__tests__/EventCreateForm.spec.js#L21-L29?

mitogh:
Thanks for your feedback and links @nealfennimore I've added most of your suggestions.

Let me know what do you think about those.

Thanks.

PR #167, opened 2018-07-18
Add tests for UI: reducer, selector and type

mitogh:
**Note** This PR requires: https://github.com/moderntribe/events-gutenberg/pull/166

### DIFF
--- a/src/modules/data/search/__tests__/actions.test.js
+++ b/src/modules/data/search/__tests__/actions.test.js
@@ -7,118 +7,50 @@ import thunk from 'redux-thunk';
 /**
  * Internal dependencies
  */
-import { actions, types } from 'data/search';
+import { actions } from 'data/search';
 
 const middlewares = [ thunk ];
 const mockStore = configureStore( middlewares );
 
 describe( '[STORE] - Search actions', () => {
 	it( 'Should add a block', () => {
-		const expectedAction = {
-			type: types.ADD_BLOCK,
-			payload: {
-				id: 1,
-			},
-		};
-		expect( actions.addBlock( 1 ) ).toEqual( expectedAction );
+		expect( actions.addBlock( 1 ) ).toMatchSnapshot();
 	} );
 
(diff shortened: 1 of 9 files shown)


src/modules/data/search/__tests__/actions.test.js
```diff
@@ -0,0 +1,147 @@
...
+
+describe( '[STORE] - Search actions', () => {
+	it( 'Should add a block', () => {
+		const expectedAction = {
+			type: types.ADD_BLOCK,
+			payload: {
+				id: 1,
+			},
+		};
+		expect( actions.addBlock( 1 ) ).toEqual( expectedAction );
```
nealfennimore: Actions are a great case for using snapshots. I'd recommend doing that with other actions so you don't have to do so much setup work. All you'd need to do is:

```js
expect( actions.addBlock( 1 ) ).toMatchSnapshot();
```

Just make sure you verify the output is correct :)

src/modules/data/ui/__tests__/reducer.test.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Internal dependencies
+ */
+import reducer, { DEFAULT_STATE } from 'data/ui/reducer';
+import { actions } from 'data/ui';
+
+describe( '[STORE] - UI reducer', () => {
+	it( 'Should return the default state', () => {
+		const current = reducer( undefined, {} );
+		expect( current ).toEqual( DEFAULT_STATE );
```
nealfennimore: Snapshots are good for reducer output as well :)

mitogh:
Thanks @nealfennimore 

Would you mind take a look at: 

- https://github.com/moderntribe/events-gutenberg/pull/166

As is one before this one and will update this one with some snapshots, thanks for the recommendation.

mitogh:
Another look on this one would be great. @nealfennimore 

Thanks for the recommendation makes test much faster to write 💃

PR #176, opened 2018-07-22
Add set of tests for Higher Order components.

mitogh:
The PR introduces tests for all the current HOC:

- `withDetails`
- `withForm`
- `withSaveData`
- `withStore`

### DIFF
--- a/src/modules/editor/hoc/__tests__/with-store.test.js
+++ b/src/modules/editor/hoc/__tests__/with-store.test.js
@@ -1,0 +1,30 @@
+/**
+ * External dependencies
+ */
+import renderer from 'react-test-renderer';
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import { withStore } from 'editor/hoc';
+
+jest.mock( 'data', () => ( {
+	getStore() {
+		return {};
+	},
+} ) );
+
+describe( 'HOC - With Store', () => {
+	it( 'Should add the store property', () => {
+		const Block = ( props ) => <div { ...props } />;
+		const Wrapper = withStore()( Block );
+		const component = renderer.create( <Wrapper /> );
+		expect( component.toJSON() ).toMatchSnapshot();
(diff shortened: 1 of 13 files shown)


src/modules/editor/hoc/with-form.js
```diff
@@ -5,13 +5,14 @@ import React, { Component } from 'react';
...
 import { connect } from 'react-redux';
+import { noop } from 'lodash';
 
 /**
  * Internal dependencies
  */
 import { actions, selectors } from 'data/forms';
 
-export default ( args ) => ( WrappedComponent ) => {
+export default ( args = noop ) => ( WrappedComponent ) => {
```
nealfennimore: I think `args` here is a misnomer. Maybe it should be `getName` or something?

Could also put a getter here like `get name` so that we isolate where `args` is used

src/modules/editor/hoc/__tests__/with-store.test.js
```diff
@@ -0,0 +1,30 @@
+/**
```
nealfennimore: What's the use case for using `withStore` instead of just using `connect` from react-redux?

src/modules/editor/hoc/with-form.js
nealfennimore: Also, can we get doc blocks here?

src/modules/editor/hoc/with-form.js
```diff
@@ -5,13 +5,14 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
```
nealfennimore: Just fyi there's a typo on this line: https://github.com/moderntribe/events-gutenberg/blob/a8e373d451d7ee1f986ced10a179688061030fc7/src/modules/editor/hoc/with-form.js#L36

mitogh:
Changes has been applied.

Thanks

PR #187, opened 2018-07-26
Add set of tests for the Form reducer

mitogh:


src/modules/data/forms/__tests__/reducer.test.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * Internal dependencies
+ */
+import reducer, { actions } from 'data/forms';
+import form, { DEFAULT_STATE } from 'data/forms/reducers/form';
+
+jest.mock( 'data/forms/reducers/form', () => {
+	const original = require.requireActual( 'data/search/reducers/search' );
+	return {
+		__esModule: true,
```
nealfennimore: Just curious why this is here? Is it because es6 imports are compiled to commonjs syntax?

mitogh: Sure.

Because we have a `default` export and not a named export on the mocked module we don't have a `default`  property to mock so by having that one we are able to mock the `default` property as a regular object.

PR #188, opened 2018-07-26
Add tests for the data/organizer reducer

mitogh:


src/modules/data/blocks/organizers/reducers/block.js
```diff
@@ -1,6 +1,9 @@
+/**
+ * Internal dependencies
+ */
 import * as types from './../types';
 
-const DEFAULT_STATE = {
+export const DEFAULT_STATE = {
 	organizer: undefined,
```
nealfennimore: Can this be replaced with `null`? JS convention is for values that will be eventually is for the value to be `null` 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null

mitogh: That's a great point, let me double check something as this one was intentionally to use `undefined` in order to prevent the block to avoid sending the value in the request. 

If this works with `null` I think it would be better, let me double check and loop back. 

Thanks

mitogh:
Changes has been applied thanks for the suggestion. 

And added some fixes to failing tests

src/modules/data/blocks/organizers/reducers/block.js
```diff
@@ -4,7 +4,7 @@
 import * as types from './../types';
 
 export const DEFAULT_STATE = {
-	organizer: undefined,
+	organizer: null,
```
nealfennimore: Indentation is off here, maybe github issue?

nealfennimore: Yeah looks like GH issue - ignore comment above

PR #189, opened 2018-07-26
Adjust code and tests to follow WP signature

mitogh:


src/modules/data/request/middlewares/__tests__/wp-request.test.js
```diff
@@ -102,7 +102,10 @@ describe( '[STORE] - wp-request middleware', () => {
 			$ajax.reject( 'Wrong path' );
 			return $ajax.promise();
 		};
-		await invoke( actions.wpRequest( { ...meta, path: 'tribe_organizer/1217//////' } ) );
+
+		try {
```
nealfennimore: Is the try block actually throwing an error?

paulmskim: same note here, is the invoke function throwing an error? we should also handle the error if there is something thrown back.

mitogh: Yes that's the case, problem is WP `wpRequest` does not have `catch` method on it where `$.Defered` does it have it. 

I think we need to move away from this one from `$` into something like: 

- https://github.com/WordPress/gutenberg/tree/master/packages/api-fetch

Or just `fetch` as we need to have access to the headers from the request and `api-fetch` just returns the response.

Let me know if you guys have any other concern here.

mitogh: Moving this one to `master` to fix the problem but let me know if you have more comments on this one so I can send another PR addressing those comments.

nealfennimore: I'm good with either @mitogh. My personal preference is that we use something that's promise based, so we can use the `await/async` syntax within the try/catch block that you have in the test.

Maybe we just throw an error within that `fail` callback in order to do that?

PR #190, opened 2018-07-26
Normalize styles for the Event Options metabox

juanfra:
🎫 https://central.tri.be/issues/110895

As required in the ticket: Normalize styles for the Event Options metabox

### DIFF
--- a/src/modules/editor/style.pcss
+++ b/src/modules/editor/style.pcss
@@ -78,3 +78,22 @@
 		box-sizing: border-box;
 	}
 }
+
+.edit-post-meta-boxes-area {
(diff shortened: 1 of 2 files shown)


src/modules/editor/style.pcss
```diff
@@ -78,3 +78,22 @@
...
+		h2 {
+			&.hndle {
+				border-bottom: none;
+				span {
+					color: #191e23;
+					font-size: 13px;
+				}
+			}
+		}
+		.handlediv {
```
mitogh: Looks good, thanks maybe just one space above to separate target elements.

PR #191, opened 2018-07-27
separate logic and presentation for event website block

paulmskim:
For ticket: https://central.tri.be/issues/108872


src/modules/blocks/event-website/container.js
```diff
@@ -0,0 +1,31 @@
...
+import EventWebsite from './template';
+
+const mapStateToProps = ( state ) => ( {
+	url: selectors.getUrl( state ),
+	urlLabel: selectors.getLabel( state ),
+} );
+
+const mapDispatchToProps = ( dispatch ) => bindActionCreators( actions, dispatch );
+
+export default withStore()(
```
mitogh: 👍

src/modules/blocks/event-website/template.js
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Internal dependencies
+ */
+import './style.pcss';
+
+const placeholder = __( 'Add Button Text', 'events-gutenberg' );
+
+const EventWebsite = ({ isSelected, url, urlLabel, setWebsite, setLabel }) => {
+
+	const renderUrlInput = () => {
```
mitogh: I think it would be great if we can move this one outside as it's own block so we can test them  more easily and where the props used inside of the function are props of the block.

src/modules/blocks/event-website/template.js
```diff
@@ -0,0 +1,109 @@
...
+					className={ inputClassNames }
+					value={ urlLabel }
+					placeholder={ placeholder }
+					onChange={ sendValue( setLabel ) }
+				/>
+			</div>
+		);
+	}
+
+	const renderPlaceholder = ( placeholder ) => {
```
mitogh: Same comments to the functions above.

src/modules/blocks/event-website/template.js
```diff
@@ -0,0 +1,109 @@
...
+				<UrlInput
+					autoFocus={ false }
+					value={ url }
+					onChange={ setWebsite }
+				/>
+			</div>
+		);
+	}
+
+	const renderLabelInput = () => {
```
mitogh: Same comments to the functions above.

src/modules/blocks/event-website/container.js
nealfennimore: Can this just be like the below?

```js
compose(
	withStore(),
	connect(
		mapStateToProps,
		mapDispatchToProps,
	),
	withSaveData(),
)( EventWebsite )
```

paulmskim: yes, done!

paulmskim:
@mitogh @nealfennimore back at you guys for another pass

src/modules/blocks/event-website/template.js
```diff
@@ -21,85 +21,82 @@ import './style.pcss';
...
+		<button className={ classNames( classes ) }>
+			{ placeholder }
+		</button>
+	);
+}
+
+const EventWebsite = ( props ) => {
+
+	const { isSelected, urlLabel } = props;
+	let eventWebsite = renderPlaceholder();
```
nealfennimore: Maybe we should update to a ternary or if/else so we're only rendering once. What do you think?

```js
const eventWebsite = isSelected || urlLabel
	? renderPlaceholder()
	: (
		[
			renderLabelInput( props ),
			renderUrlInput( props ),
		]
	)

```

paulmskim: i'm down with this.

PR #194, opened 2018-07-27
Update tests to prevent errors on snapshots

mitogh:


nealfennimore:
Try cherry picking f110778. That should fix the security error

PR #196, opened 2018-07-30
[Sharing Links] Pass the attributes to the view

juanfra:
🎫 https://central.tri.be/issues/109050


mitogh:
_removing this branch_

cc @juanfra

PR #198, opened 2018-07-31
Use DBA format for dates when saving into the store

mitogh:
In order to allow backwards compatibility we need to set the dates in the same format used on the PHP side

### DIFF
--- a/src/modules/blocks/classic-event-details/block.js
+++ b/src/modules/blocks/classic-event-details/block.js
@@ -44,10 +44,11 @@ import { actions as websiteActions, selectors as websiteSelectors } from 'data/b
 import { actions as classicActions, selectors as classicSelectors } from 'data/blocks/classic';
 import {
 	actions as organizersActions,
-	selectors as organizerSelectors
+	selectors as organizerSelectors,
 } from 'data/blocks/organizers';
 import { actions as UIActions } from 'data/ui';
 import { sendValue } from 'editor/utils/input';
+import { FORMATS } from 'editor/utils';
 
 /**
  * Module Code
@@ -122,7 +123,7 @@ class EventDetails extends Component {
 					placeholder={ __( 'Organizer', 'events-gutenberg' ) }
 					onChange={ sendValue( setOrganizerTitle ) }
 				/>
-				<EventOrganizers store={ store } />
+				<EventOrganizers store={ store }/>
 			</MetaGroup>
 		);
 	}
@@ -143,9 +144,14 @@ class EventDetails extends Component {
 		const { start, toggleDashboardDateTime } = this.props;
 
 		return (
-			<div onClick={ toggleDashboardDateTime }>
-				<strong>{ __( 'Start: ', 'events-gutenberg' ) }</strong><br/>
-				{ toDate( toMoment( start ) ) }
+			<div
+				role="button"
+				tabIndex={ 0 }
+				onKeyDown={ toggleDashboardDateTime }
+				onClick={ toggleDashboardDateTime }
+			>
+				<strong>{ __( 'Start: ', 'events-gutenberg' ) }</strong><br />
+				{ toDate( toMoment( start ), FORMATS.WP.date ) }
 				{ this.renderStartTime() }
 			</div>
 		);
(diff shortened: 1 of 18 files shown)


src/modules/blocks/classic-event-details/block.js
```diff
@@ -143,9 +144,14 @@ class EventDetails extends Component {
 		const { start, toggleDashboardDateTime } = this.props;
 
 		return (
-			<div onClick={ toggleDashboardDateTime }>
-				<strong>{ __( 'Start: ', 'events-gutenberg' ) }</strong><br/>
-				{ toDate( toMoment( start ) ) }
+			<div
```
paulmskim: Could we use a `button` element here instead? This would allow us to get rid of the `role`, `tabIndex`, and `onKeyDown` properties. Maybe we don't need to do it in this PR, but there's one thing to look at that may be unexpected here:

- with the `onKeyDown` property, if a user is focused on the `div` by tabbing and presses any key (doesn't have to be `enter` or `space`), it'll trigger the action. I don't think this is what is expected.

src/modules/blocks/classic-event-details/block.js
```diff
@@ -170,8 +176,8 @@ class EventDetails extends Component {
 		const { end, toggleDashboardDateTime } = this.props;
 		return (
 			<div onClick={ toggleDashboardDateTime }>
```
paulmskim: I'm guessing the same would apply here too.

src/modules/blocks/classic-event-details/block.js
mitogh: Thanks for flagging this; we have another `branch` that changes how the inside the `event-details` handles this, I think a `button` might be the best option.

However in the other comment not sure if is semantically correctly as contains more nested `div` and `<h2>` so I think in the other case might be better to use the `role` attribute instead.

nealfennimore: +1 for button tags. Ideally, we would only want [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) within a button. @mitogh could you update it so the content does not contain divs/h2s?

mitogh: Sure. In cases where the inner content has a `input` or other complex structure I think we should use a `role` right?

Asking for the `branch` that uses the inner as `input` instead of just text.

nealfennimore: Can you link where this is done? I can comment on it there :)

mitogh: - https://github.com/moderntribe/events-gutenberg/blob/103745/feature-human-readable-input/src/modules/elements/date-input/element.js#L84

nealfennimore: @mitogh That role should be fine there if it's intended to not have valid content inside. I did see that it only looks to have a `h2` used with it: https://github.com/moderntribe/events-gutenberg/commit/3179e704d4afe6ab0ccacf37f6e46493a44bd79f#diff-2eea6f66c0a9b23134499cbd10bd412bR306

Would it make more sense to make the `h2` a `span`? I'm not sure what else you're planning on using it for, but browsers already handle `button` tags in an accessible manner, so I would suggest changing the `h2`.

mitogh: I think a `<span>` might be enough for that one however I think the problem is on the component itself as it renders an `<input>`.

- https://github.com/moderntribe/events-gutenberg/blob/103745/feature-human-readable-input/src/modules/elements/date-input/element.js#L63

nealfennimore: `input` is fine - it's within the MDN docs as valid for button https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content :sunglasses:

nealfennimore: Let's keep it as it is then @mitogh. I don't think we'll get around this unless we change the component around

paulmskim: Ideally in that case, you could have either the children in a button show (if not selected) or the input show (if selected).

mitogh: Great, will update accordingly on the next PR once is open, thanks for the feedback.

As those changes are not part of this branch.

Thanks

PR #199, opened 2018-07-31
Feature/109049 fe view for date time

paulmskim:
For ticket: https://central.tri.be/issues/109049


src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,97 @@
+<?php
+
+if ( ! function_exists( 'tribe_is_same_day' ) ) {
```
juanfra: Can we add docblocks for each function? I think it would be a good idea to move these to a template-tags file

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,97 @@
...
+		$date_2_year = date( 'Y', strtotime( $date_2 ) );
+		return $date_1_year == $date_2_year;
+	}
+}
+
+if ( ! function_exists( 'tribe_get_event_date_time' ) ) {
+	function tribe_get_event_date_time( $date, $date_format, $time_format ) {
+		$timestamp = strtotime( $date );
+
+		return [
```
juanfra: I assume once it's merged we'll need to support the same PHP version as TEC. There are plans to bump the number. But we need to be sure that everything we introduce here is compatible with that number (I think by the end of the year it'll end up being 5.6)

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,97 @@
...
+			'date' => date( $date_format, $timestamp ),
+			'time' => date( $time_format, $timestamp ),
+		];
+	}
+}
+
+$event_id = get_the_ID();
+
+$start = get_post_meta( $event_id, '_EventStartDate' )[0];
```
juanfra: perhaps nesting to make the code mode readable

src/views/blocks/event-datetime.php
paulmskim: Sure, I'm not sure what the preferred approach here is for functions such as these, maybe @mitogh or @bordoni can weigh in.

src/views/blocks/event-datetime.php
paulmskim: how do you mean?

paulmskim: ah you mean lining up the left / right. gotcha.

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,97 @@
...
+$end = get_post_meta( $event_id, '_EventEndDate' )[0];
+$timezone = get_post_meta( $event_id, '_EventTimezone' )[0];
+$separator_date = get_post_meta( $event_id, '_EventDateTimeSeparator' )[0];
+$separator_time = get_post_meta( $event_id, '_EventTimeRangeSeparator' )[0];
```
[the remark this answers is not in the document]
paulmskim: yeah, i didn't see that til you mentioned it. will update.

src/views/blocks/event-datetime.php
mitogh: I don't think we have a defined way for doing this at the moment, so I think having a `common`, or `utils` directory would be helpful.

Another thing here, besides folder structure and more in terms of the code itself is about `strtotime` it might be better to use something that takes the timezone into account such as: 

- https://github.com/moderntribe/tribe-common/blob/master/src/Tribe/Date_Utils.php#L568

src/views/blocks/event-datetime.php
bordoni: Yeah lets use `array()`, it's ugly but required for now.

src/views/blocks/event-datetime.php
paulmskim: @bordoni @mitogh I don't know how you want to set this up / hook it up, so I'm going to leave it in this file for now. Could one of you set this up or give me? Or you can give me clear instructions on where it should live, what I should call it, and how to link it into the current system.

paulmskim:
@juanfra I've made the changes you've requested. Only thing that's left is the common or utils file, which I have no idea how the best way to set it up is. I've left a note for Crisoforo and Gustavo.

juanfra:
Thanks a lot @paulmskim! Looking great 🎉

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+if ( ! function_exists( 'tribe_get_event_date_time' ) ) {
+	function tribe_get_event_date_time( $date, $date_format, $time_format ) {
+		$timestamp = wp_strtotime( $date );
+
+		return array(
+			'date' => date( $date_format, $timestamp ),
+			'time' => date( $time_format, $timestamp ),
+		);
+	}
+}
```
bordoni: We need to move this into the correct place, an external file, we don't want to mix definition of methods with Templates.

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+
+$event_id = get_the_ID();
```
bordoni: Lets use `$this->get( 'post_id' )`

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+
+$start          = get_post_meta( $event_id, '_EventStartDate', true );
+$end            = get_post_meta( $event_id, '_EventEndDate', true );
+$timezone       = get_post_meta( $event_id, '_EventTimezone', true );
+$separator_date = get_post_meta( $event_id, '_EventDateTimeSeparator', true );
+$separator_time = get_post_meta( $event_id, '_EventTimeRangeSeparator', true );
+$all_day        = (boolean) get_post_meta( $event_id, '_EventAllDay', true );
```
bordoni: Lets use `tribe_is_truthy()`

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+$same_day       = tribe_is_same_day( $start, $end );
+$show_year      = ! tribe_is_same_year( $start, $end ) || ! tribe_is_same_year( $start, date( 'Y-m-d H:i:s' ) );
+
+$date_format_with_year = 'F j, Y';
```
bordoni: We need to use the Format from the Settings page.

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+$date_format_no_year   = 'F j';
```
bordoni: Same for this.

src/views/blocks/event-datetime.php
```diff
@@ -1,7 +1,122 @@
...
+$time_format           = 'g:i a';
```
bordoni: Same for this.

src/views/blocks/event-datetime.php
paulmskim: definitely. could you either do this for me, as I don't know how you want it set up, or give me clear directions on where you'd want it and how to hook it up into the existing code?

src/views/blocks/event-datetime.php
paulmskim: I tried this, but it always returned `null`. Is there something that's not hooked up correctly?

bordoni: That is weird, it was working in the past. We need to add as a Default Arg into every block. On the render method of every block before `add_template_globals` we should have a `$args['post_id']` to set which post we are working with.

paulmskim: so does that mean I just add `$args['post_id'] = get_the_ID();`? Or is there another way to get the post ID that you'd like to use?

PR #200, opened 2018-07-31
Make the price work with two decimals.

juanfra:
🎫 https://central.tri.be/issues/111154

### DIFF
--- a/src/modules/editor/utils/range.js
+++ b/src/modules/editor/utils/range.js
@@ -23,12 +23,13 @@ export function parser( input ) {
 	}
 
 	const [ a, b ] = extractParts( chars );
+	const [ num_a, num_b ] = [ parseFloat( a ), parseFloat( b ) ];
 
-	if ( ! b || b === a ) {
-		return a === 0 ? '' : trim( a );
+	if ( ! num_b || num_b === num_a ) {
+		return num_a === 0 ? '' : trim( a );
 	}
 
-	return a >= b ? `${ trim( b ) } - ${ trim( a ) }` : `${ trim( a ) } - ${ trim( b ) }`;
+	return num_a >= num_b ? `${ trim( b ) } - ${ trim( a ) }` : `${ trim( a ) } - ${ trim( b ) }`;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/modules/editor/utils/range.js
```diff
@@ -65,7 +65,7 @@ export function extractParts( chars ) {
 			return null === result ? '' : result[ 1 ];
 		} )
 		.filter( ( item ) => ! isEmpty( item ) )
-		.map( ( item ) => parseFloat( item ) )
+		.map( ( item ) => parseFloat( item ).toFixed(2) )
```
mitogh: I think this one should be placed after the `filter` operation as it might be `NaN` here.

bordoni:
There are still tests failing @juanfra

juanfra:
Yes, the problem here is that the `toFixed()` function returns a string. But at the same time I got some feedback from QA/Strategy and they actually want this to work as following (i.e.):

1. 45.00 it should be 45.00
2. 45 it should be 45
3. 45.123123123 should be 45.12
4. 45.5 should be 45.50

So, I need to re-work the patch.

juanfra:
I finally had to re-engineer it to make it respect if the users insert values with decimals (even values like `2.00`). In order to do that, `extractParts` is now returning an array of strings instead of numbers.

src/modules/editor/utils/range.js
```diff
@@ -65,7 +66,10 @@ export function extractParts( chars ) {
 			return null === result ? '' : result[ 1 ];
 		} )
 		.filter( ( item ) => ! isEmpty( item ) )
-		.map( ( item ) => parseFloat( item ).toFixed( 2 ) )
+		.map( ( item ) => {
+			// If the user input the price with decimals (even .00) we want to keep them
+			return ( 0 < item.indexOf( '.' ) ) ? parseFloat( item ).toFixed( 2 ) : parseFloat( item ).toFixed( 0 );
```
mitogh: Seems like the only difference on the condition is the amount of decimals I think refactor this into a variable will make it more clear, for instance:

```js
const decimals = 0 < item.indexOf( '.' ) ? 2 : 0 ;
return parseFloat( item ).toFixed( decimals );
```

juanfra:
@mitogh I still need your approval to merge this one. Thanks buddy!

PR #203, opened 2018-08-01
Update `resources` structure

mitogh:


bordoni:
That is something that we want to keep the structure like we have on the actual plugins. Which `src` is the root of any code

mitogh:
Okay, so we should close this one and consider the audit complete @bordoni ?

paulmskim:
@mitogh I think that we still have a some things left from the audit that we haven't completed:

- There are still a lot of blocks and elements which need logic and presentation separated. I've done the website and price blocks, and most recently the time picker element, but I think that's about all we have at the moment.
- I'm still seeing `setInitialState` functions in blocks, actions, and reducers which I think we were supposed to remove.
- I may have missed a discussion somewhere, but I don't think the third point on the improvements list was addressed yet either.

mitogh:
Thanks @paulmskim , I think I should give an update after the call, however here is a summary.

1. Yes, this is an ongoing process so we keep things moving so whenever we have a chance to split a block should favor stateless.
1. Yes, we are keeping those because of the nature of Gutenberg and also as is handled automatically by the withSaveData HOC
1. Same as previous point here, we are using a pattern called “[normalize state shape](
https://egghead.io/lessons/javascript-redux-normalizing-the-state-shape)” ,  somafter the call we agreed to keep it.
1. Tests, by now we have the store mostly tested with útiles functions and same as first point is an ongoing process for the next PRs requiring tests and make sure old ones are passing

So in summary the only one that is not done or in process is the one addressed by this PR, so it seems like we are not changing this one so we should be complete on the audit.

Let me know if you have any additional questions.

bordoni:
@paulmskim lets take this conversation to Slack and if needed check what tickets needs to be created to address the problems above. I would also like @nealfennimore  to participate.

For the time been I am closing this specific PR, but it doesn't mean that we addressed all the audit questions and requests.

mitogh:
_removing this branch_

PR #204, opened 2018-08-01
Fix/110215 datetime time picker issue

paulmskim:
For ticket: https://central.tri.be/issues/110215

This was only supposed to be to allow `Multi Day` to be used with `All Day` on both time pickers, but it turned out to be a much deeper thing than anticipated. Here's a recap:

- The time picker elements had a great deal of logic, which was interfering with being able to set both time pickers to `All Day` with `Multi Day`. All the logic in the time picker was moved out to the `DateTime` block.
- With setting times / dates, `All Day`, and `Multi Day`, there was a host of extra overwriting that was happening through middlewares that was also interfering with the initial task. All this logic was moved to the relevant functions and all block middlewares were removed.
- The date time actions, types, and reducer were cleaned up to remove unused actions, types, and reducer cases.

Summary of what's been done here:

- All block middlewares have been removed.
- Time picker element has become stateless.
- All logic for middlewares and time picker element has moved to the date time block.
- Extra actions, types, and reducer cases have been removed.

One thing that is left is to fix the broken test, which I find extremely weird why it's failing. @mitogh could you possibly have a look?

Lastly, @mitogh I recognize that this work may conflict with some things you're doing for the human readable input ticket, so if you can keep an eye out for things and we can resolve them when it comes merge time, that'd be great.


paulmskim:
Hey Crisoforo, so I remember discussing with you a few weeks back on my thoughts for having middlewares for one specific action, and this is how we can do this without using any middlewares. Basically, the main reason for moving all the logic out of the middlewares was because there were a lot of side effects happening. If I tried to change the multiday toggle, for example, it would change the all-day setting, or vice versa. Times would also be changed to depending on the configuration.

All of the logic could have happened using pure function and reducers rather than using middleware and having multiple side effects for one action. If we pared down the middlewares to target only the intended action, then it begs the question for the need for middleware in the first place.

This is also part of separating logic and presentation, which hasn't completely happened yet for the Date Time block, but I was able to do for the Time Picker. Once we can move the logic to its own file and feed in properties to the presentational component, it'll be significantly cleaner.

I feel like we could get into a discussion about this, so I think the easiest thing would be to hop onto a call and work it out. I'd love input from @nealfennimore as well on this, and maybe could join us for a chat.

mitogh:
I think having this one as a `thunk` actions would make more sense if we want to stay away from `middlewares`, however I don't think we should tho, as my main concern here is being able to fire a single action from an external component and from that point on rely on the middleware to handle the adjustment of other settings.

Let's have a call to discuss if `middlewares` or `actions` but I think all of this logic is part of the "business logic" should be testable and outside of any component and available to other actions as we need to interact with other stuff in future such as: 

- Recurrence
- Community Events

That uses / rely on this "business logic" without having to be explicit part of this block.

paulmskim:
After discussions with @mitogh and @nealfennimore we've decided to keep logic in event handlers and create separate utility functions encapsulating logic at a later date when extension of logic to various parts of the application is required.

src/modules/blocks/event-datetime/block.js
```diff
@@ -51,17 +52,21 @@ import {
...
+	toDateTime,
+	isSameDay,
+	replaceDate,
 	toTime,
 } from 'utils/moment';
 import { FORMATS, timezonesAsSelectData, TODAY } from 'utils/date';
 import { HALF_HOUR_IN_SECONDS } from 'utils/time';
 import withSaveData from 'editor/hoc/with-save-data';
 import { hasClass, searchParent } from 'editor/utils/dom';
+import { DAY_IN_SECONDS, HOUR_IN_SECONDS } from '../../editor/utils/time';
```
mitogh: We should use the alias here `editor/utils` or `utils/time` for consistency and make it more easier to move in future if required.

paulmskim:
hey @mitogh so I agree with all your points above, we definitely should move the logic into its own functions. As mentioned in slack, I'd like to do the refactoring of the logic into thunks in another PR as I'd like to get this fix merged in first for the release. Also, I believe there are already tests for the actions, there were no new actions added.

mitogh:
Seems like the fix for some failing tests is the issue with timezones, I did an adjustment on tests on this PR: https://github.com/moderntribe/events-gutenberg/pull/207 however I haven't merge that one yet as will make this one to require an extra change so wanted to ask if you are okay with moving that one into `master` first.

On the other hand the tests for the actions just ensure the actions have the desired signature, but the tests that ensures the logic for core functionally works as expected was part of the middlewares, and now by moving those into this block we don't have tests for those.

I would love to keep this one on `hold` until we have the `tests` in place in order to prevent any problem from upcoming changes by not having tests on the core part of the extension.

I'm able to provide some help on this one if you want as I'm going to have some space today.

src/modules/blocks/event-datetime/block.js
```diff
@@ -10,6 +10,7 @@ import { compose, bindActionCreators } from 'redux';
  * WordPress dependencies
  */
 import { Component } from '@wordpress/element';
+import { applyFilters } from '@wordpress/hooks';
```
bordoni: Are we still using this?

src/modules/blocks/event-datetime/block.js
```diff
@@ -52,17 +53,21 @@ import {
...
+	toDateTime,
+	isSameDay,
+	replaceDate,
 	toTime,
 } from 'utils/moment';
 import { FORMATS, timezonesAsSelectData, TODAY } from 'utils/date';
 import { HALF_HOUR_IN_SECONDS } from 'utils/time';
 import withSaveData from 'editor/hoc/with-save-data';
 import { hasClass, searchParent } from 'editor/utils/dom';
+import { DAY_IN_SECONDS, HOUR_IN_SECONDS } from '../../editor/utils/time';
```
bordoni: I would like for us to not use `../../` aka relative paths.

mitogh:
Should we change the base of this PR to point to the `develop` branch @paulmskim ?

Thanks

src/modules/blocks/event-datetime/block.js
paulmskim: This was done previously and I hadn't touched it. I will remove all instances of it.

src/modules/blocks/event-datetime/block.js
paulmskim: Yes we are. Was it removed from the code before?

src/modules/data/blocks/datetime/__tests__/__snapshots__/reducers.test.js.snap
```diff
@@ -36,18 +36,6 @@ Object {
 }
 `;
 
-exports[`[STORE] - Datetime reducer Should set the end time 1`] = `
```
mitogh: We should test all the elements of the reducer such as: 

- Start
- End
- Multidday 
- All day

src/modules/data/blocks/datetime/__tests__/thunks.test.js
```diff
@@ -0,0 +1,182 @@
...
+		expect( store.getActions() ).toMatchSnapshot();
+	} );
+
+	test( 'Set end time', () => {
+		const attributes = {
+			end: '2018-06-05 17:30:00',
+			seconds: 64800,
+		};
+
+		store.dispatch( thunks.setStartTime( attributes ) );
```
mitogh: Wrong `thunk` called. I think is `setEndTime`

src/modules/data/blocks/datetime/__tests__/thunks.test.js
```diff
@@ -0,0 +1,182 @@
...
+	} );
+
+	test( 'Set all day on', () => {
+		const attributes = {
+			start: '2018-06-05 17:00:00',
+			end: '2018-06-05 17:30:00',
+			isAllDay: true,
+		};
+
+		store.dispatch( thunks.setAllDayThunk( attributes ) );
```
mitogh: Having the `Thunk` suffix doesn't seem needed as the method is exported already with the namespace. `thunks.setAllDay` it seems like should be enough.

src/modules/data/blocks/datetime/__tests__/thunks.test.js
```diff
@@ -0,0 +1,182 @@
...
+		store.dispatch( thunks.setDates( attributes ) );
+
+		expect( store.getActions() ).toMatchSnapshot();
+	} );
+
+	test( 'Set multi day to true', () => {
+		const attributes = {
+			start: '2018-06-05 17:00:00',
+			end: '2018-06-05 17:30:00',
+			checked: true,
```
mitogh: Wonder what this `checked` value means?

src/modules/data/blocks/datetime/__tests__/thunks.test.js
```diff
@@ -0,0 +1,182 @@
...
+		};
+
+		store.dispatch( thunks.setMultiDayThunk( attributes ) );
```
mitogh: Same as above with the `Thunk` suffix

src/modules/data/blocks/datetime/__tests__/thunks.test.js
mitogh: If is the `multiday` value I think `isMultiDay` as a name would give a better context around this one.

mitogh:
Couple of questions / comments.

src/modules/data/blocks/datetime/__tests__/__snapshots__/reducers.test.js.snap
paulmskim: These snapshots were removed from removing the tests for the actions that set the time. These have since been moved to thunks.

src/modules/data/blocks/datetime/__tests__/thunks.test.js
paulmskim: Yes, you're right! Will fix.

src/modules/data/blocks/datetime/__tests__/__snapshots__/reducers.test.js.snap
mitogh: Right, however we still need to test the reducers besides the test on the thunks.

src/modules/data/blocks/datetime/__tests__/thunks.test.js
paulmskim: done

src/modules/data/blocks/datetime/__tests__/__snapshots__/reducers.test.js.snap
paulmskim: This reducer had all the tests you had written in it to start. The only thing that happened was actions were removed and so were their tests. Is there something else that needs to be done here? I'm not sure what more I am supposed to do.

paulmskim:
@mitogh I'll change the base of this PR to the `develop` branch once the above changes are approved. The `develop` branch and this branch are very similar, so the diff won't show much at all, as I based the `develop` branch off of this.

mitogh:
Great, I think we are missing just couple of things here @paulmskim 

1. Test the  missing reducers not the `thunks`, as it seems that the tests for the `thunks` is completed.

- https://github.com/moderntribe/events-gutenberg/pull/204/files#diff-9a8285a808259152919c79ca9b2759f1

1. Merge `master` into the `develop` branch.
1. Point this PR into that branch so we can approve and we merge this branch there.

From there I will update my PR #207 to adjust to the new `thunks` instead of middleware and code.

src/modules/data/blocks/datetime/__tests__/reducers.test.js
```diff
@@ -23,14 +23,6 @@ describe( '[STORE] - Datetime reducer', () => {
 		expect( reducer( DEFAULT_STATE, actions.setEnd( 'June 25, 2018 4:00 pm' ) ) ).toMatchSnapshot();
 	} );
 
-	it( 'Should set the start time', () => {
```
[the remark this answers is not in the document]
paulmskim: @mitogh See below:

- https://github.com/moderntribe/events-gutenberg/pull/204/files/b7b8f41c2afaa711a3d7ec383ff17f0916e7490a#diff-22dc3bb2b61e847036a0573950fb5e7eL10
- https://github.com/moderntribe/events-gutenberg/pull/204/files/b7b8f41c2afaa711a3d7ec383ff17f0916e7490a#diff-22dc3bb2b61e847036a0573950fb5e7eL32
- https://github.com/moderntribe/events-gutenberg/pull/204/files/b7b8f41c2afaa711a3d7ec383ff17f0916e7490a#diff-22dc3bb2b61e847036a0573950fb5e7eL39
- https://github.com/moderntribe/events-gutenberg/pull/204/files/b7b8f41c2afaa711a3d7ec383ff17f0916e7490a#diff-22dc3bb2b61e847036a0573950fb5e7eL39

These actions were removed. I cannot test actions that do not exist. I've tested the functions that have been moved in the thunks. Could you check the actions and make sure the tests line up (including the ones that are removed)?

mitogh:
It seems like `develop/datetime-and-human-readable-input` was based from this PR not `master` as the PR diff has only the changes from the merge from `master` I think we should be able to have the whole `diff` .

mitogh:
Looks like we need @bordoni approval here.

bordoni:
Go for it, but the tests seems to be broken.

mitogh:
We have a fix on the human branch, and should be fixed before create the final PR pointing to `master`

PR #206, opened 2018-08-02
Feature/108086 upsell support

paulmskim:
For ticket: https://central.tri.be/issues/108086

### DIFF
--- a/src/modules/blocks/event-datetime/block.js
+++ b/src/modules/blocks/event-datetime/block.js
@@ -329,7 +329,7 @@ class EventDateTime extends Component {
 								{ this.renderMultidayToggle() }
 							</div>
 						</div>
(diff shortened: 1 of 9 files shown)


src/modules/blocks/event-datetime/block.js
```diff
@@ -492,6 +495,34 @@ class EventDateTime extends Component {
 		);
 	}
 
+	renderUpsell = () => {
```
mitogh: I think this one might be a good case to have a state less component so we can start reducing the complexity of this block.

Even on this same file but as a separate component would be fine I guess something along the lines: 

```js
const UpShell = () => {
}
```

paulmskim: yeah, fair enough. hold up!

PR #207, opened 2018-08-02
Human readable input

mitogh:
Introduce human readable input.

Ticket: https://central.tri.be/issues/103745


src/modules/elements/date-input/element.js
```diff
@@ -0,0 +1,107 @@
...
+		children: null,
+		value: '',
+		onClickHandler: noop,
+		setDateTime: noop,
+		setNaturalLanguageLabel: noop,
+	};
+
+	constructor( props ) {
+		super( props );
+		this.inputRef = createRef();
```
paulmskim: The ref is defined here and set in the input element below, but never used anywhere. I don't think there's a need for this if we're not actually using the ref.

src/modules/elements/date-input/element.js
```diff
@@ -0,0 +1,107 @@
...
+		this.sendDateTime = debounce( this._sendDateTime, 250 );
```
paulmskim: We don't need this to be defined here, it should just be called in `handleChange()`. I'm mainly pointing this out because without the ref and this, we can get rid of the constructor and the need to use a stateful component.

src/modules/elements/date-input/element.js
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	handleChange = ( event ) => {
```
paulmskim: this doesn't have to be done in this PR, but would love to move this logic out to the component or block that is actually using this. that way this element, if we wanted to, could be more reusable.

src/modules/elements/date-input/element.js
```diff
@@ -0,0 +1,107 @@
...
+	renderInput() {
+		const { value, onClickHandler } = this.props;
+		return (
+			<input
+				type="text"
+				name="date-input"
+				className="tribe-editor__date-input"
+				value={ value }
+				ref={ this.inputRef }
+				onBlur={ this.onBlur }
```
paulmskim: An `onBlur` function hasn't been defined in this component.

src/modules/elements/date-input/element.js
```diff
@@ -0,0 +1,107 @@
...
+				onChange={ this.handleChange }
+				onFocus={ onClickHandler }
```
paulmskim: if we can handle the onChange just like this, that'd be best (see note on `onChange` function above).

src/modules/elements/date-input/style.pcss
```diff
@@ -0,0 +1,9 @@
+.tribe-editor__date-input {
+	width: 100%;
+
+	input[type="text"] {
+		width: 100%;
+		//border: none;
```
paulmskim: remove?

src/modules/elements/date-input/element.js
mitogh: Right, should be removed, let me do that thanks for flagging this one.

src/modules/elements/date-input/element.js
mitogh: However if we do defined in the change the reference is going to be created every time the `handleChange` is fired by having this one in the constructor we just have a single definition and a single `debounce` call.

paulmskim: that's right, my bad. so instead, once this becomes stateless, we could just do something like: `const sendDateTimeDebounce = debounce( this._sendDateTime, 250 );` and call that. I'm fine with this staying, but once we move this to stateless, that's how it'll be defined.

mitogh:
Would you mind take another look here @bordoni @nealfennimore @paulmskim, thanks

PR #213, opened 2018-08-04
Initial folder structure

nealfennimore:
- Adds in initial folder structure for multiple plugins
- Separate out jest config and setup to run multiple projects
- Add in nvmrc file with node version from SQ1
- Remove aliasing from jest and add in babel based aliasing
- Add in jsconfig.json for IDE module lookup

@bordoni @mitogh @paulmskim I've changed aliasing as we would not be able to point to files within separate plugins with the aliasing we had now. It's been added in as `@@plugins/events` (double `@` to not conflict with npm modules) for anything in the `plugins/events/src/modules` dir.

I'm fine with changing it to something else, but we should have a discussion if we want it to be this way (it could also be something like `@@events`).

_*NOTE:* `@@plugins` has been updated to `@@tribe`_

## Folder Structure

```
repo
    ├── plugins
        ├── common
        ├── events
        ├── tickets
        ├── tickets-plus
        ├── events-pro
    ├── other folders
    ├── files to the global 
```

## Needs feedback
- plugin aliasing: `@@plugins/events`
- yarn vs npm - yarn.lock was gitignored so I un-ignored it. Should we remove package-lock.json if we're preferring yarn?

## TODO:
- Get webpack working with builds
- Fix eslint plugin alias (currently linter errors on valid imports, will come back to later. Put as warning in linter for time being)


.editorconfig
```diff
@@ -1,2 +1,21 @@
-[{*.js}]
-indent_style=tab
+root = true
```
mitogh: ❤️

.editorconfig
```diff
@@ -1,2 +1,21 @@
...
+
+[*.{css,pcss}]
+indent_style = tab
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Matches the exact files either package.json or .travis.yml
+[{package.json,.travis.yml}]
```
mitogh: Can we use the `*` here as well for any `.json` or `.yml` file added in future like mock data or similar.

.babelrc
```diff
@@ -19,6 +19,12 @@
 		} ]
 	],
 	"plugins": [
+		["module-resolver", {
+			"root": ".",
+			"alias": {
+				"@@plugins/events": "./plugins/events/src/modules"
```
mitogh: Is this one TEC?

plugins/events/src/modules/blocks/classic-event-details/block.js
```diff
@@ -14,7 +14,7 @@ import AutosizeInput from 'react-input-autosize';
 import { __ } from '@wordpress/i18n';
 import { Component } from '@wordpress/element';
 
-import withSaveData from 'editor/hoc/with-save-data';
+import withSaveData from '@@plugins/events/editor/hoc/with-save-data';
```
mitogh: I think this one should be part of `common` however I'm not sure if we are moving everything over here and then into `common`

.babelrc
nealfennimore: yes I'll put the folder structure in the PR

.editorconfig
nealfennimore: sure can :sunglasses:

plugins/events/src/modules/blocks/classic-event-details/block.js
mitogh: Sure I think all the current HOC so they are available across the rest of the plugins.

nealfennimore: Cool, added HOC folder in. Might need to be structured like the events folder, but not sure if should have something like the `src/modules` dirs

bordoni:
@nealfennimore I think it would be great to call it `@@tribe/` to be ours.

bordoni:
About the Yarn vs NPM, I like Yarn Better but we have had problems with it on Gutenberg core builds, let's talk Monday about it.

mitogh:
In terms of YARN vs NPM I think `npm` should be good enough, in terms of that also wanted to flag if we should take advantage of NPM orgs instead so we can prefix and push each stuff (later into his own pacakge) so we don't have to worry about conflicts with namespaces and we can use `@tribe/common` instead of `@@tribe/common` also in this way each package can be deploy under the NPM orgs.

- https://www.npmjs.com/docs/orgs/

mitogh:
Also another question is should be start flagging all the elements that are part of `common` or we can move it later?

nealfennimore:
I'll start handling the common stuff after I get builds working again. 

For the NPM discussion, I started aliasing with the double `@@` for that exact reason. We already have a `@moderntribe` npm organization, so we'd probably just transition `@@tribe` to `@moderntribe` eventually. I'm going to keep it at `@@` as I want to avoid naming conflicts/lookups with npm, while retaining a similar syntax.

As for NPM vs yarn... I'm disliking npm a lot when it comes to package management (i.e. having to `npm uninstall --save-dev [package]` vs `yarn remove [package]` ). Yarn makes the CLI a lot easier and less verbose and avoids having to remember to include arguments that I find superfluous.

The yarn not working for builds is worth looking into. After I get builds working again, let's see if yarn is working, otherwise we should switch to npm

bordoni:
I would go with Yarn too, NPM sometimes takes to long. That sounds good Neal, let's go with Yarn.

What would be the difference between using `@@tribe` and `@moderntribe` in terms of using the NPM organization?

nealfennimore:
Right now `@@tribe` is aliased internally to be compiled with babel. Eventually, if we migrate to uploading the packages to npm, we could import a plugin like so `yarn add @moderntribe/events`.

Meaning all we would have to do would be to replace `@@tribe` to `@moderntribe`, and npm would handle imports within the codebase. But this would mean publishing every plugin to npm

mitogh:
Should we remove the branch from the PR @nealfennimore ?

nealfennimore:
Deleting it now

mitogh:
:awthanks:

plugins/common/src/modules/hoc/__tests__/with-details.test.js
```diff
@@ -9,7 +9,7 @@ import thunk from 'redux-thunk';
 /**
  * Internal dependencies
  */
-import { withDetails } from 'editor/hoc';
+import { withDetails } from '@@tribe/common/hoc';
```
paulmskim: @nealfennimore I think this actually needs to be `@@tribe/common/editor/hoc` as `@@tribe/common` only points to `plugins/common/src/modules`

plugins/common/src/modules/hoc/__tests__/with-form.test.js
```diff
@@ -9,7 +9,7 @@ import thunk from 'redux-thunk';
 /**
  * Internal dependencies
  */
-import { withForm } from 'editor/hoc';
+import { withForm } from '@@tribe/common/hoc';
```
paulmskim: Same thing here

plugins/common/src/modules/hoc/__tests__/with-save-data.test.js
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 /**
  * Internal dependencies
  */
-import { withSaveData } from 'editor/hoc';
+import { withSaveData } from '@@tribe/common/hoc';
```
paulmskim: same here

plugins/common/src/modules/hoc/__tests__/with-store.test.js
```diff
@@ -7,9 +7,9 @@ import React from 'react';
 /**
  * Internal dependencies
  */
-import { withStore } from 'editor/hoc';
+import { withStore } from '@@tribe/common/hoc';
```
paulmskim: same here

plugins/events/src/modules/blocks/classic-event-details/index.js
```diff
@@ -12,7 +12,7 @@ import { __ } from '@wordpress/i18n';
  */
 import EventDetails from './block';
 import { Icons } from 'elements';
-import { withStore } from 'editor/hoc';
+import { withStore } from '@@tribe/common/hoc';
```
paulmskim: same thing here

plugins/common/src/modules/hoc/__tests__/with-details.test.js
nealfennimore: Had to so some fancy symlinking here to get it to work that way. The more updated aliasing is done here: https://github.com/moderntribe/events-gutenberg/pull/218/commits/0a440d79c98ca2a057f13735d68e37410c875586

PR #215, opened 2018-08-06
Fix/111884 event links separation

paulmskim:
For ticket: https://central.tri.be/issues/111884


src/modules/blocks/event-links/template.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const renderButtons = ( props ) => (
+	<div key="event-links" className="tribe-editor__block tribe-editor__events-link">
+		{ renderGoogleCalendar( props ) }
+		{ renderiCal( props ) }
+	</div>
+);
+
+const renderControls = ({
```
mitogh: If this is a component should be capitalized and I think `render` prefix might not be required as can be consider as a component. 

(Ref)[https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized]

As this is a rule for JSx only, I think it would be great if we follow this pattern regardless in order to introduce another common pattern.

src/modules/blocks/event-links/template.js
```diff
@@ -0,0 +1,136 @@
...
+				className="tribe-editor__btn-input"
+				value={ iCalLabel }
+				placeholder={ iCalExportPlaceholder }
+				onChange={ sendValue( setiCalLabel ) }
+			/>
+		</div>
+	);
+};
+
+const renderButtons = ( props ) => (
```
mitogh: Same as comment below for the name.

src/modules/blocks/event-links/template.js
```diff
@@ -0,0 +1,136 @@
...
+				className="tribe-editor__btn-input"
+				value={ googleCalendarLabel }
+				placeholder={ googleCalendarPlaceholder }
+				onChange={ sendValue( setGoogleCalendarLabel ) }
+			/>
+		</div>
+	);
+};
+
+const renderiCal = ({
```
mitogh: Can this be a component as well if so, to apply the same pattern for the name.

src/modules/blocks/event-links/template.js
```diff
@@ -0,0 +1,136 @@
...
+const iCalExportPlaceholder = __( 'iCal Export', 'events-gutenberg' );
+
+const renderPlaceholder= ( label ) => (
+	<button className="tribe-editor__btn--link tribe-editor__btn--placeholder" disabled>
+		<LinkIcon />
+		{ label }
+	</button>
+);
+
+const renderGoogleCalendar = ({
```
mitogh: Same as below for naming.

src/modules/blocks/event-links/template.js
```diff
@@ -0,0 +1,136 @@
...
+import './style.pcss';
+
+/**
+ * Module Code
+ */
+
+const googleCalendarPlaceholder = __( 'Google Calendar', 'events-gutenberg' );
+const iCalExportPlaceholder = __( 'iCal Export', 'events-gutenberg' );
+
+const renderPlaceholder= ( label ) => (
```
mitogh: Same as below as well for naming.

src/modules/blocks/event-links/template.js
paulmskim: Hey, so we did not do this in the previous 2 separation of concerns (the price block and website block). There was a comment about separating the sub-rendering functions out of the block itself (https://github.com/moderntribe/events-gutenberg/pull/191#pullrequestreview-141155091 and https://github.com/moderntribe/events-gutenberg/blob/ef424fd4c92eb6d86d8237353356adecff7764c8/src/modules/blocks/event-price/template.js) but we did not create them into their own components nor capitalize each function.

The above link you shared (https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized) is for components that are exported and used elsewhere in the code (as they may clash with built-in tags). If we wanted to reuse it, then I would move it into its own element and allow it to be exported. However I would not want to create all of these functions into their own components, it just doesn't make sense given their purpose.

I personally think this isn't a key issue in this PR, so maybe we can move on without worrying about this. Thoughts on this?

mitogh: I think is not a major issue, however my concern is based on the following.

1. Having a function that behaves and has the exact same signature of a stateless component, it can be categorized as a component and in order to avoid confusion I think we should name and use as components otherwise having signature and body of a stateless component and used them as function is a bit confusing.

1. Compostability, I think this one is not a huge part as those components are only used on this file, however by consider the functions components is easier to use patterns to compose components for example: 

```js
<div key="event-links" className="tribe-editor__block tribe-editor__events-link">
	{ renderGoogleCalendar( props ) }
	{ renderiCal( props ) }
</div>
```

We can compose into something like: 

```js
<div key="event-links" className="tribe-editor__block tribe-editor__events-link">
	<ButtonLink
		classname="tribe-events-gcal"
		name="google-calendar-label"
		value={ googleCalendarLabel }
		usePlaceholder={ ! hasGoogleCalendar && ! hasiCal }
		placeholder={ googleCalendarPlaceholder }
		onChange={ sendValue( setGoogleCalendarLabel ) }
	 />
	<ButtonLink
		classname="tribe-event-ical"
		name="tribe-event-ical"
		value={ iCalLabel }
		usePlaceholder={ ! hasGoogleCalendar && ! hasiCal }
		placeholder={ iCalExportPlaceholder }
		onChange={ sendValue( setiCalLabel ) }
	 />
</div>
```

Having the opportunity to rewrite this files allow us to use components to remove functions that behaves very similar.


Lets see what everyone else thinks. 👍

mitogh: cc @nealfennimore & @bordoni

nealfennimore: I'm fine with keeping it as is. Though it's true that it _could_ be a component, if we already have everything working, and this was taken out of a class structure in order to modularize things, we can easily transition to proper components later when time permits

PR #217, opened 2018-08-06
Feature: Add Edit Link button for Venue and Organizer block

mitogh:


src/Tribe/Editor.php
```diff
@@ -368,6 +372,10 @@ public function assets() {
 			array(
 				'in_footer'    => false,
 				'localize'     => array(
+					array(
+						'name' => 'tribe_js_config',
+						'data' => apply_filters( 'tribe_events_gutenberg_js_config', $js_config ),
```
bordoni: Docblock.

src/modules/elements/edit-link/element.js
```diff
@@ -0,0 +1,53 @@
...
+			href={ `${ admin }post.php?post=${ id }&action=edit` }
+			target={ target }
+			{ ...extraProps }
+		>
+			{ label }
+		</a>
+	);
+};
+
+EditLink.propTypes = {
```
bordoni: Why not use `static` call? inside of the class? I think I added that to our Babel.

mitogh: Mainly because they are a stateless component they don't have any of that type of mechanism (`class` associated) because they are functions.

PR #218, opened 2018-08-06
Lerna working builds

nealfennimore:
- Adds in lerna for building between projects
- Abstract out webpack configs for reuse
- Update some [plugins](https://github.com/moderntribe/events-gutenberg/commit/4158517000d6c9857498ab4ed71df8f9fab53cdb)
- Add in webpack and tribe common plugin to global webpack configs
- Add in analyze, dev, and build scripts


webpack/utils/directories.js
```diff
@@ -0,0 +1,14 @@
+const { lstatSync, readdirSync } = require( 'fs' );
+const { join } = require( 'path' );
+
+const isDirectory = source => lstatSync( source ).isDirectory();
+const getDirectories = source => (
```
mitogh: Wonder why not inline export instead. `export const ...`

webpack/module/rules/styles.js
```diff
@@ -0,0 +1,35 @@
+const MiniCssExtractPlugin = require( 'mini-css-extract-plugin' );
+
+const plugin = new MiniCssExtractPlugin( {
```
mitogh: Same question about the inline export.

webpack/externals/tribe/common.js
```diff
@@ -0,0 +1,6 @@
+const pickBy = require( 'lodash/pickBy' );
```
mitogh: How about `import { pickBy } form 'lodash'`

webpack/utils/directories.js
nealfennimore: No this needs to be in `commonjs` syntax - we're not compiling ES6 modules down to CJS syntax before webpack can use it

mitogh: Oh! There was a reason 😆 .

Thanks for the heads up.

webpack/externals/tribe/common.js
nealfennimore: Not necessary in this case, lodash is already modularized and this is part of the build process

nealfennimore:
@bordoni @mitogh I made a couple changes:
- Updated to use `npm` instead of `yarn`... `yarn` was causing issues with workspaces  and how that was working in conjunction with lerna.
- `@@tribe` is now `@moderntribe`
- `window.tribe.common.hoc` now exists and is properly bundled away from `window.tribe.events`. I now need this file `hoc.js` in the asset pipeline as the events plugin depends on it. 
- `common` plugin has some dependencies on `events` plugin and that can be removed now
- Added in an `analyze` script so you guys can visually see the bundle dependencies (we should probably add a vendor bundle, there's a lot of duplicate vendor modules)

bordoni:
I like those changes, my only problem is, do we need to release stuff into NPM? For `@moderntribe`?

PR #222, opened 2018-08-07
Styles front and back

nealfennimore:
Adds in webpack config for style extraction for backend and frontend styles


mitogh:
I see, asking because now the BE has some FE so `FE` is no longer just the `FE` views is also part of the BE 😆

mitogh:
:turkey-dance:

mitogh:
Shall we remove this branch as well @nealfennimore ?

nealfennimore:
It's :boom:

PR #223, opened 2018-08-08
Move PHP and Correct files + Missing JS conflicts

bordoni:
_Ref:_ [C#110788](http://central.tri.be/issues/110788)
---

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
-composer.lock -diff
+*.lock -diff
 package-lock.json -diff
-yarn.lock -diff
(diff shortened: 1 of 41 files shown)


bordoni:
@mitogh just added that.

.gitattributes
```diff
@@ -0,0 +1,4 @@
+composer.lock -diff
+package-lock.json -diff
+yarn.lock -diff
```
mitogh: Having a regex would make it more simpler I think
`*.lock`

PR #224, opened 2018-08-08
Datetime & Human Readable issues

mitogh:
PR with the changes from @paulmskim & myself merged into a single PR.


src/modules/elements/time-picker/element.js
```diff
@@ -184,86 +116,69 @@ export default class TimePicker extends Component {
...
+	const renderDropdownContent = ({ onClose }) => (
+		<ScrollTo>
+			{ () => (
+				<ScrollArea
+					id="tribe-element-timepicker-items"
+					key="tribe-element-timepicker-items"
+					role="menu"
+					className={ classNames( 'tribe-editor__timepicker__items' ) }
+				>
+					{ renderItem( { text: 'All Day', value: 'all-day' }, onClose ) }
```
bordoni: Translate `All Day`

src/modules/elements/time-picker/element.js
```diff
@@ -184,86 +116,69 @@ export default class TimePicker extends Component {
...
 }
+
+TimePicker.defaultProps = {
+	current: moment(),
+	min: undefined,
+	max: undefined,
+	start: undefined,
+	end: undefined,
+	step: HALF_HOUR_IN_SECONDS,
+	timeFormat: 'H:i',
```
bordoni: Shouldn't we bring that from the defaults picked by the user, on the Settings page?

mitogh: Yes, this is provided via props, however if not provided we have a default value.

mitogh: - https://github.com/moderntribe/events-gutenberg/blob/c3a8f33a5ce713c79c8b7eb41fc7fff0d024cf62/src/modules/blocks/event-datetime/block.js#L475

- https://github.com/moderntribe/events-gutenberg/blob/c3a8f33a5ce713c79c8b7eb41fc7fff0d024cf62/src/modules/blocks/event-datetime/block.js#L513

PR #225, opened 2018-08-08
Add editable timezone on date label

mitogh:
*Ticket*: https://central.tri.be/issues/110991


src/modules/data/utils.js
```diff
@@ -8,7 +13,11 @@
  */
 export const maybeDispatch = ( attributes, dispatch ) => ( action, key, defaultValue ) => {
 	if ( key in attributes ) {
-		dispatch( action( attributes[ key ] || defaultValue ) );
+		const useDefault = isUndefined( attributes[ key ] ) ||
```
nealfennimore: Can this be extended to all falsy values, or is it necessary for just these cases?

src/modules/elements/timezone/__tests__/element.test.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { mount } from 'enzyme';
```
nealfennimore: Feel free to set this in jest globally: https://github.com/moderntribe/calendar-saas-app/blob/release/1.17.3/jest.setup.js#L40-L42

src/modules/data/utils.js
mitogh: In this case we still want to use `false` as we might have toggle UI elements that have `true` as default but we want to use  the saved value such as `false`   or other values like `0` so we have to make it a bit custom

mitogh:
Might require another set of 👀

src/modules/data/utils.js
nealfennimore: That's what I figured but wanted to make sure :+1:

PR #227, opened 2018-08-09
Fix/111885 event organizer separation

paulmskim:
For tickets:
- https://central.tri.be/issues/111885
- https://central.tri.be/issues/111898

Notes:
- I ended up doing both tickets in one as they were somewhat related to each other
- I changed around a number of variables, naming, and other patterns mainly for consistency and to be able to follow the code easier. I had quite a difficult time figuring out what does what, especially around the organizer block, and for any new devs that may have to hop onto this project at a later date, I think this is important that the code is easy to read and follow.
- I fixed up the compose and id -> clientId warnings.

I expect there will be discussions about some of the changes, and I would love for us to open the conversation to having a more unified convention and making things easier to follow.


src/modules/blocks/event-organizer/container.js
```diff
@@ -0,0 +1,93 @@
...
+
+const onCreateNew = ( ownProps ) => ( title ) => (
+	ownProps.createDraft( {
+		title: {
+			rendered: title,
+		},
+	} )
+);
+
+const mapStateToProps = ( state, props ) => ( {
```
nealfennimore: Minor thing, but can we change `props` to `ownProps`, like you have in the `mergeProps` function?

src/modules/data/search/reducer.js
```diff
@@ -0,0 +1,90 @@
+/**
```
mitogh: For this would like to stick to the "reducer composition pattern" to keep reducer more simpler.

src/modules/data/details/__tests__/types.test.js
```diff
@@ -5,8 +5,8 @@ import { types } from 'data/details';
 
 describe( '[STORE] - Details types', () => {
 	it( 'Should return the types values', () => {
-		expect( types.SET_DETAILS_POST_TYPE ).toBe( 'SET_DETAILS_POST_TYPE' );
-		expect( types.SET_DETAILS_LOADING ).toBe( 'SET_DETAILS_LOADING' );
+		expect( types.SET_POST_TYPE ).toBe( 'SET_POST_TYPE' );
+		expect( types.SET_IS_LOADING ).toBe( 'SET_IS_LOADING' );
```
mitogh: Would like to keep this one and the search with the prefix to quickly identify the "loading" state into the debugger, otherwise `SET_IS_LOADING` can be fired from multiple places

src/modules/blocks/event-organizer/container.js
paulmskim: this was just copied over from the old file, but definitely will do :+1:

src/modules/data/blocks/index.js
```diff
@@ -1,6 +1,6 @@
 /**
  * Internal dependencies
  */
-import reducer from './reducers';
+import reducer from './reducer';
```
mitogh: Would like to keep the plural name for those reducers that combine other reducers to quickly identify where a group of reducer exists.

src/modules/blocks/event-organizer/container.js
```diff
@@ -0,0 +1,93 @@
...
+		onCreateNew: onCreateNew( ownProps ),
+	};
+};
+
+export default compose(
+	withStore( { isolated: true, postType: ORGANIZER } ),
+	withForm( ( props ) => props.clientId ),
+	connect( mapStateToProps ),
+	withDetails( 'organizer' ),
+	connect( null, mapDispatchToProps, mergeProps ),
```
mitogh: Why do we have to connect twice?

mitogh:
Seems like we do have some failing tests as well, would like to fix those before the merge.

src/modules/data/details/__tests__/types.test.js
paulmskim: Sure, then the actions would need those prefixes too. What we had here was an action with type of `SET_DETAILS_LOADING` but the action creator called `setLoading()`. This makes the code difficult to follow, especially like you said where we have multiple places firing `SET_IS_LOADING`.

mitogh: Right! I think using the same pattern from the type into the action creator will help with that, I think we can add "alias" into during the binding to make it more easier to use in the props if required but agree consistency should be favor in this case

src/modules/data/search/reducer.js
paulmskim: So the reducer composition pattern is used to split logic between reducers (for example, one action adds an item to the state, another filters items in the state, another maps items in the state, etc.). In this case, it makes sense to handle the logic of the actual item separately from the addition, mapping, and filtering. What we had before was we were just passing data through one reducer into another, there wasn't any separation of logic.

src/modules/data/blocks/index.js
paulmskim: So I'm actually going to disagree with this here, we do not need a file name to tell us whether we have a combined reducer or a single reducer. The root reducer is a combined reducer, but we are not calling it reducers, we are calling it the root reducer (and `reducer.js`). In this case, I feel consistency is more important than having some file names singular and some file names plural.

src/modules/blocks/event-organizer/container.js
paulmskim: So this is something I want to discuss with you and @nealfennimore . The `withDetails()` HOC takes in an id, in this case organizer, which maps to the prop and gets the organizer id. However, in order to do that, `withDetails()` needs to be passed in the state with the organizer id as a prop. I don't think creating this dependency on state to prop from `withDetails()` is the way to compose HOCs together, it's basically a hierarchical requirement. Also, in order to successfully remove the logic from presentational, because much of the logic uses props passed down from the HOCs, the `connect()` should be the last in the `compose()` function. However, it's proving very difficult to decouple these properly. This was a quick solution to fix this, but it requires more work to dig into how to decouple these effectively.

src/modules/data/search/reducer.js
mitogh: Not logic, it also provides other things like: 

- keep the reducers as flat as possible, this helps to add new cases into the reducer
- easier to test, same as before as is a flat structure you don't need to worry to much during the test phase

In this case does not seems like major difference as is just avoid the need to write an array of data on every reducer, however this pattern allow for flexibility in terms of search for items later the road as you can store the IDs of the object in a separate array to manage stuff more granular. 

Some reference: https://egghead.io/lessons/react-redux-reducer-composition-with-objects

src/modules/data/blocks/index.js
nealfennimore: How about importing it as `rootReducer` or `initialReducer(s)`?

mitogh: I see your point here and I think it makes sense to favor consistency as well so lets keep it as it's as in the end returns a single object

src/modules/blocks/event-organizer/container.js
mitogh: Great would like to have a TODO here with a path on the fix if is not solved on this PR.

`withDetails` gets a key as a prop, so this key is where the ID of an object is stored in the state of the HOC the HOC does not passed the  `ID` into the wrapped component the only 2 things passed are: 

- Details are being fetched - `bool`
- Details - `object`

I'm not sure if I get this part

> needs to be passed in the state with the organizer id as a prop. I don't think creating this dependency on state to prop from withDetails()

src/modules/data/details/__tests__/types.test.js
paulmskim: Fixing this now

src/modules/blocks/event-organizer/container.js
paulmskim: So the HOCs are meant to be composable, meaning they're basically plug-and-play. In this case, the `withDetails` HOC is dependent on mapping the state -> organizer id -> pass in as a prop to the component before the `withDetails` HOC. This dependency requires that the organizer id in the state mapped to the prop is a hierarchical requirement of this HOC to work properly rather than allowing the HOC to be composable.

mitogh: I see your point and I think is right ideally you should be able to place HOC on any order of precedence, I think this one might be a good fit to convert into a SAGA to fetch the details instead of a HOC

mitogh:
I think we just need to add two things into this PR to move it: 

- https://github.com/moderntribe/events-gutenberg/pull/227#discussion_r209061696
- Add `// TODO` comment with the path to remove the multiple `connect` calls

PR #231, opened 2018-08-10
Fix some linter issues

mitogh:


mitogh:
Fixes some present issues on the lint, still a bunch more to address but can be a good start point.

src/modules/data/blocks/venue/__tests__/utils.test.js
```diff
@@ -44,7 +44,7 @@ describe( 'Venue Utils', () => {
 
 		const expected = {
 			lat: 37.802953,
-			lng: -122.448342,
+			lng: - 122.448342,
```
paulmskim: Is this actually a lint issue? This is the first time I've seen a negative number with a space between the `-` and the number.

src/modules/elements/venue-form/element.js
```diff
@@ -179,7 +179,9 @@ export default class VenueForm extends Component {
 					tagName="h3"
 					format="string"
 					value={ title }
-					onChange={ ( value ) => { this.setState( { title: value } ) } }
+					onChange={ ( value ) => {
+						this.setState( { title: value } );
+					} }
```
paulmskim: not necessary to fix, but as a personal preference I would rather just return the function rather than opening a body to invoke a single function:
`onChange={ ( value ) => this.setState( { title: value } ) }`
this is fine though if you don't want to adjust this.

src/modules/data/blocks/venue/__tests__/utils.test.js
mitogh: great observation will update lint rules to avoid this on negative numbers, didn't realize as it was an automatic fix 🤦‍♂️

src/modules/elements/venue-form/element.js
mitogh: Do you know if there's any rule for this so we can update the linter?

paulmskim: I think this was because we had the `{ }` already set. The `return` statement is a totally different command from just invoking the function, so I don't think there's anything automatic for this. We'll just need to fix it ourselves when we see it.

mitogh: Seems like I was able to find one: 

- https://eslint.org/docs/rules/arrow-body-style

So it should be flagged now 🎉

PR #233, opened 2018-08-10
Feature/111175 price block default currency symbol

paulmskim:
For ticket: https://central.tri.be/issues/111175

@bordoni would love your input on the PHP side for this, I got it working the way I wanted it to but I'm not sure if this is how you want to organize the code.

Also a note for later (will add a ticket for this), there are already various settings that are being passed from the PHP side to the window object. It would be great if we could pass a single object that looks something like:

```
window.tribe_blocks_editor_settings = {
  editor: { ... },
  blocks: {
    price: { ... }
    ...
  },
  ...
};
```

or some other method of organizing the settings.


mitogh:
I think something Sam suggested is great I've created a single filterable object which is named exactly same as the one used on agency.

`tribe_js_config` I think it would be great if all config are stored in that one, I've added the comment in the ticket created so we can follow up from there.

src/Tribe/Editor.php
```diff
@@ -280,6 +280,18 @@ public function is_events_post_type() {
...
 	}
 
+	/**
+	 * Check if current admin page is new post page
+	 *
+	 * @since  0.2.6-alpha
+	 *
+	 * @return bool
+	 */
+	public function is_new_post() {
```
bordoni: We don't need to add a new Method `tribe( 'context' )->is_new_post()` will give you that.

bordoni:
@mitogh I wanted to get this merged and talk about what that settings field will look like after the architecture.

mitogh:
Yep just flagging this but we have a ticket to follow up later on on that one. @bordoni

mitogh:
We need to solve conflicts and tests here tho.

src/modules/data/blocks/price/actions.js
```diff
@@ -37,8 +39,16 @@ export const setDescription = ( description ) => ( {
 } );
 
 export const setInitialState = ( { get } ) => ( dispatch ) => {
+	const isNewEvent = isTruthy( getPriceSettings().is_new_event );
```
mitogh: I think changing this value directly into the reducer will remove the need for the PHP check as the default state is used only when this value is empty.

- https://github.com/moderntribe/events-gutenberg/blob/c6bbbafcb533e7d070f2e16566d05f1ce2fea198/src/modules/data/blocks/price/reducers.js#L8-L9

Also that will remove the need for this function as we can just assign ` getPriceSettings().default_currency_symbol` directly into the default state

paulmskim: So because `setInitialState()` runs after `DEFAULT_STATE` is set, it's overridden with blank strings. This is also the trouble of having a setInitialState() function that overrides the default state, because everything needs to be done inside this function. I've created a ticket to look into how we can remove this.

mitogh: I see, I'm not sure why uses the white space here, tho but also be aware that that this function might be called by `event-datetime` as well as it also render this block on it so that might be the problem on the other hand.

There's a function `maybeDispatch` that can be replaced with `dispath` here to use a default value if the original value is undefined or empty string.

paulmskim: As this should only work on a new event, that won't work as intended. It still needs to check the `isNewEvent` variable, then apply the correct value.

paulmskim: For example, if the user actually removes the currency symbol, saves the event, then refreshes, they will unexpectedly see a $ symbol there.

mitogh: In the case of use `maybeDispatch`, however `get` function return the value that is present or saved on the attributes if is not present on the attributes list returns a default value instead.

mitogh:
Seems tests are failing on this one.

PR #237, opened 2018-08-13
Render the timezone based on editor settings

mitogh:


src/views/blocks/event-datetime.php
```diff
@@ -89,6 +92,8 @@
 			</span>
 		<?php endif; ?>
 
-		<span class="tribe-events-schedule__timezone"><?php echo $timezone; ?></span>
+        <?php if ( $show_time_zone ) : ?>
+            <span class="tribe-events-schedule__timezone"><?php echo esc_html( $time_zone_label ); ?></span>
+        <?php endif; ?>
```
paulmskim: I think the file is using tabs, but you've got spaces here?

PR #247, opened 2018-08-15
common store setup

mitogh:
Couple of things to consider on this PR, also some left questions for discussion where I would like some input.

- `request` middleware was copied over the one that is present in `events` it was not moved in order to make sure things are still working after the merge in the branch but eventually once all blocks are migrated to use `common` store is going to be removed from `events` and eventually this one should use a native promised based mechanism for the request `fetch` (we have a ticket for this already, fetch implementation).


plugins/common/src/modules/store/configureStore.js
```diff
@@ -0,0 +1,34 @@
...
+import thunk from 'redux-thunk';
+
+/**
+ * Internal dependencies
+ */
+import reducer from '@moderntribe/common/data';
+import { wpRequest } from './middlewares';
+
+export default () => {
+	let store = window.__tribe_common_store__ || null;
```
mitogh: Q: Should we define a convention for `private` / `internal` / `shared` global variables?

plugins/common/src/modules/data/reducers/plugins/reducer.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { types } from '@moderntribe/common/data/reducers/plugins';
+
+export default ( state = {}, action ) => {
+	switch ( action.type ) {
+		case types.ADD_PLUGIN:
+		case types.DEACTIVATE_PLUGIN:
```
mitogh: Q: Should we remove `ACTIVATE/DEACTIVATE` cases as I'm not sure if we should have any differences between those cases and `ADD/REMOVE`. 

Q: I think this might be a follow up of the previous question, as if we don't need `ACTIVATE/DEACTIVATE` we can just use `true/false` values for the `ADD/REMOVE` of the plugin instead of removing the entry from the reducer.

plugins/common/src/modules/store/configureStore.js
nealfennimore: `private` variables tends to be prefixed with a `_` by convention. Is there a variable you can point to that you think is `shared`?

mitogh: I think this variable is shared but via the proxy method `getStore`, however is still reachable via the `window` object.

plugins/common/src/modules/store/index.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Internal dependencies
+ */
+import configureStore from './configureStore';
+
+const store = configureStore();
+
+export const getStore = () => store;
```
nealfennimore: maybe export `store` too? I know we need `getStore` as it's used by that one HOC, but there's no reason `store` cannot be exported as well

plugins/common/src/modules/data/reducers/plugins/reducer.js
nealfennimore: I'm not sure whether we'd be ever activating or deactivating a plugin, so I'm not sure that we need this. Maybe @bordoni  would know of some cases where we would need to do this

plugins/common/src/modules/store/index.js
mitogh: How about updating the `HOC` to use the reference `store` and remove `getStore`?

nealfennimore: Even better :)

mitogh: Will do 🕺 ! thanks for the suggestion

plugins/common/src/modules/store/configureStore.js
nealfennimore: okay that's fine `__` is commonly used for items on the window so :+1:

plugins/common/src/modules/data/reducers/plugins/types.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Internal dependencies
+ */
+import { withPrefix } from '@moderntribe/common/data/utils';
+
+export const ADD_PLUGIN = withPrefix( 'ADD_PLUGIN' );
+export const REMOVE_PLUGIN = withPrefix( 'REMOVE_PLUGIN' );
+export const ACTIVATE_PLUGIN = withPrefix( 'ACTIVATE_PLUGIN' );
+export const DEACTIVATE_PLUGIN = withPrefix( 'DEACTIVATE_PLUGIN' );
```
paulmskim: I'm being picky here, but could we do `${ PREFIX }/ADD_PLUGIN` here instead and not use `withPrefix()`? Though this is a simple example, it makes things more difficult to follow and I feel it's not necessary to have a function to prefix an action.

I feel that we may define more constants in the future, we could move the `PREFIX` to a `constants.js` file in the `data` folder.

plugins/common/src/modules/store/configureStore.js
paulmskim: not a big issue here, but the `null` here isn't needed, since if `__tribe_common_store__` is not in the `window` object it'll return undefined.

plugins/common/src/modules/data/reducers/plugins/types.js
nealfennimore: Agreed, can we keep it like we have here? https://github.com/moderntribe/events-gutenberg/pull/247/files/cbd53515c483c5fadd1b39d7d7180d4f311a52c0#diff-84532eecdca8c10d24d07af96fe146a3R9

paulmskim: One more thing, actually, is I'm not sure if we'll need to use another prefix, or if we'll have a prefix for other use cases (not just for separating plugins like `@@MT/COMMON`), but it might be prudent to use `COMMON_PREFIX` instead so that if we need to prefix something else, we won't have to refactor and it becomes easier to follow.

mitogh: Sounds good, however I'm not sure on the last comment. You are saying something along the following: 

- `@@MT/PREFIX_COMMON/ACTION`
- `@@MT/COMMON_PREFIX/ACTION`

I think that makes sense where `PREFIX` is specific to the reducer / action like: 


- `@@MT/COMMON_REQUEST/ACTION_NAME`

plugins/common/src/modules/store/configureStore.js
mitogh: Right, however I'm using `null` instead to indicate the lack of an object instead as pointed here by Neal. 

- https://github.com/moderntribe/events-gutenberg/pull/188#discussion_r205663142

plugins/common/src/modules/data/reducers/plugins/types.js
mitogh: Great I think we already have that on in place 🕺

plugins/common/src/modules/store/configureStore.js
nealfennimore: `null` ends up being the default value here instead of stating the intention that this value will eventually be changed.

In this context, it would make more sense to remove the `or` statement after `window.__tribe_common_store__`, as if `__tribe_common_store__` is `undefined`, then the value would already be falsey, and all proceeding logic would remain the same. So the `null` here a bit redundant :)

mitogh:
Changes applied as suggested.

Thanks.

plugins/common/src/modules/data/reducers/plugins/reducer.js
mitogh: Hey @bordoni would you mind to take a look over here ⬆️.

Thanks

plugins/common/src/modules/data/utils.js
```diff
@@ -0,0 +1 @@
+export const PREFIX = '@@MT/COMMON';
```
paulmskim: this is still `PREFIX` here, we should have `COMMON_PREFIX` instead and export that to use everywhere.

paulmskim:
Just one thing left, and travis failing. Once those are addressed this looks good to go!

plugins/common/src/modules/data/utils.js
mitogh: Not sure if I get this one, as  this is just inside of `common` so there no need to add that prefix as is going to be the only one on this directory as every plugin has his own directory.

paulmskim: So `prefix` is a very generic word, it could be used in a variety of contexts, and in the future we may want to prefix other parts of the code. If we just use `PREFIX` here, then later add something like `ORGANIZER_PREFIX`, we or new developers won't easily know what `PREFIX` is as it's a very generic meaning. If we, instead, start with `COMMON_PREFIX`, then later find the need to add `ORGANIZER_PREFIX`, this makes it easy to do.

mitogh:
Answer, rebuilding travis is failing in some occasions.

plugins/common/src/modules/data/utils.js
mitogh: I don't think I have any opinions as you can identify the source of the variable via the module name where it comes from, however I see the benefit as it removes the need to export the variable with alias where is used on the same file. 

So will update. 

I'm going to use `PREFIX_COMMON_STORE` to make it a bit clear, tho.

mitogh:
Mind take another look @paulmskim ?

paulmskim:
should we add a changelog entry here?

mitogh:
Yes thanks for the reminder.

plugins/common/src/modules/data/reducers/plugins/reducer.js
bordoni: I feel like we might want to keep these are they are different scenarios in general. But I don't think we would ever hit the React APP on remove, deactivate, activate or add.

PR #252, opened 2018-08-16
Add events plugin into the `common` store

mitogh:
The following PR address the following tickets on central:

- https://central.tri.be/issues/112009
- https://central.tri.be/issues/112008
- https://central.tri.be/issues/112007
- https://central.tri.be/issues/112006
- https://central.tri.be/issues/112005
- https://central.tri.be/issues/112004
- https://central.tri.be/issues/112001

This one is not required and can be decliend as is not part of the store: 

- https://central.tri.be/issues/112010
- https://central.tri.be/issues/112002


mitogh:
Hey!

This PR contains all the migration of the store into common in order to have the blocks working at all times we needed to complete this PR with all the changes so that's why I moved a head with this one.

One thing I would like to ask for discussion is the namespace used by this plugin, based on the doc TEC is using `events` namespace but I'm not 100% about that, however I think we should set a guidelines for `TEC` and the rest of plugins to make it consistent with the current applications sites I think `events` is usually called `TEC` so wanted to flag this one before move a head with the merge. 

Thanks.

package.json
```diff
@@ -16,7 +16,7 @@
 		"npm": ">=5.0.0"
 	},
 	"scripts": {
-		"test": "TZ=UTC jest --config jest.config.json",
+		"test": "TZ=UTC jest --config ./jest.config.json",
```
nealfennimore: Was this not working before for you?

paulmskim:
should we add a changelog entry here?

mitogh:
Yes 👍

mitogh:
Updated this PR with some follow up commits so if you can take another look before I move a head with the merge would be great.

Thanks @paulmskim @nealfennimore @bordoni

plugins/common/src/modules/data/reducers/plugins/reducer.js
```diff
@@ -11,18 +11,12 @@ import { types } from '@moderntribe/common/data/reducers/plugins';
 export default ( state = {}, action ) => {
 	switch ( action.type ) {
 		case types.ADD_PLUGIN:
```
paulmskim: A thought on this here, since we only have `ADD_PLUGIN` and `REMOVE_PLUGIN`, and as the only state for the key `action.payload.name` is `true`, wouldn't it make sense instead to have the state as an array that holds the active plugins instead? This seems to simplify this model quite a bit.

mitogh: Right, this will work if this is know before hand, however this is going to be controlled by each plugin instead so reducers here are attached 1 by one once the bundle of the specific plugin loads.

Example: https://github.com/moderntribe/events-gutenberg/blob/7e8603a1335e9dabd9d412716081ce333ceb2100/plugins/events/src/modules/data/index.js#L9-L14

So this one is going to be located into each plugin isolated.  

I think in case something like `array` is required to be injected here can be always be abstracted into an action creator, thunk or saga _eventually_

mitogh:
Changes has been applied.

PR #266, opened 2018-08-23
Add the RSVP confirmation functionality

juanfra:
🎫 https://central.tri.be/issues/112195

### DIFF
--- a/src/Tribe/Blocks/Rsvp.php
+++ b/src/Tribe/Blocks/Rsvp.php
@@ -197,7 +197,7 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 		 *
 		 * @param $data $_POST Parameters comes from RSVP Form
 		 */
-		do_action( 'tribe_tickets_rsvp_before_order_processing', $_POST );
+		do_action( 'tribe_tickets_rsvp_before_order_processing' );
 
 		$attendee_details = tribe( 'tickets.rsvp' )->parse_attendee_details();
 
--- a/src/Tribe/Blocks/Rsvp.php
+++ b/src/Tribe/Blocks/Rsvp.php
@@ -205,8 +205,10 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 			wp_send_json_error( $response );
 		}
 
+		$products = (array) tribe_get_request_var( 'product_id' );
+
 		// Iterate over each product
-		foreach ( (array) $_POST['product_id'] as $product_id ) {
+		foreach ( $products as $product_id ) {
 			if ( ! $ticket_qty = tribe( 'tickets.rsvp' )->parse_ticket_quantity( $product_id ) ) {
 				// if there were no RSVP tickets for the product added to the cart, continue
 				continue;
--- a/src/Tribe/Blocks/Rsvp.php
+++ b/src/Tribe/Blocks/Rsvp.php
@@ -252,7 +252,11 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 			 * @param string $attendee_order_status 'yes' if the user indicated they will attend
 			 */
 			$send_mail_stati = apply_filters(
-				'tribe_tickets_rsvp_send_mail_stati', $send_mail_stati, $order_id, $post_id, $attendee_order_status
+				'tribe_tickets_rsvp_send_mail_stati',
+				$send_mail_stati,
+				$order_id,
+				$post_id,
+				$attendee_order_status
 			);
 
 			// No point sending tickets if their current intention is not to attend
--- a/src/Tribe/Blocks/Rsvp.php
+++ b/src/Tribe/Blocks/Rsvp.php
@@ -13,9 +13,9 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 		// Add AJAX calls
 		add_action( 'wp_ajax_rsvp-form', array( $this, 'rsvp_form' ) );
 		add_action( 'wp_ajax_nopriv_rsvp-form', array( $this, 'rsvp_form' ) );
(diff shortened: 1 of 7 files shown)


src/Tribe/Blocks/Rsvp.php
```diff
@@ -170,4 +172,112 @@ public function rsvp_form() {
...
+		do_action( 'tribe_tickets_rsvp_before_order_processing', $_POST );
+
+		$attendee_details = tribe( 'tickets.rsvp' )->parse_attendee_details();
+
+		if ( false === $attendee_details ) {
+			wp_send_json_error( $response );
+		}
+
+		// Iterate over each product
+		foreach ( (array) $_POST['product_id'] as $product_id ) {
```
bordoni: Lets use `tribe_get_request_var( 'product_id' )` to fetch this variable.

src/Tribe/Blocks/Rsvp.php
```diff
@@ -170,4 +172,112 @@ public function rsvp_form() {
...
+		$has_tickets = false;
+		$post_id     = get_the_id();
+		$ticket      = tribe( 'tickets.rsvp' )->get_ticket( $post_id, $ticket_id );
+
+		/**
+		 * RSVP specific action fired just before a RSVP-driven attendee tickets for an order are generated
+		 *
+		 * @param $data $_POST Parameters comes from RSVP Form
+		 */
+		do_action( 'tribe_tickets_rsvp_before_order_processing', $_POST );
```
bordoni: We dont need to pass `$_POST`, since it's a global we just have the hook.

src/Tribe/Blocks/Rsvp.php
```diff
@@ -170,4 +172,112 @@ public function rsvp_form() {
...
+			 *
+			 * Any attendee order status not listed here will trigger a non attendance email.
+			 *
+			 * @param array  $send_mail_stati       An array of default stati triggering an attendance email.
+			 * @param int    $order_id              ID of the RSVP order
+			 * @param int    $post_id               ID of the post the order was placed for
+			 * @param string $attendee_order_status 'yes' if the user indicated they will attend
+			 */
+			$send_mail_stati = apply_filters(
+				'tribe_tickets_rsvp_send_mail_stati', $send_mail_stati, $order_id, $post_id, $attendee_order_status
```
bordoni: break it into multiple lines.

src/resources/js/rsvp.js
```diff
@@ -1,37 +1,185 @@
 (function( $, my ) {
 	'use strict';
 
-	var tribe_tickets_rsvp = {
-		num_attendees: 0,
-		event        : {}
-	};
+	// base elements
+	var body       = $( 'html, body' );
+	var document   = $( document );
```
paulmskim: Is there a reason for `body` and `document` to be defined here? I don't see it being used in this file.

src/resources/js/rsvp.js
```diff
@@ -1,37 +1,185 @@
...
+	 * @param obj button The dom object of the clicked button
+	 * @return void
+	 */
+	function tribe_rsvp_toggle_actions( button ) {
 
-	} );
+		// Check if is the going or not going button
+		var going   = button.hasClass( 'tribe-block__rsvp__status-button--going' );
+		var sibling = going ? '.tribe-block__rsvp__status-button--not-going' : '.tribe-block__rsvp__status-button--going';
+		var sibling = button.closest( '.tribe-block__rsvp__status' ).find( sibling );
```
paulmskim: If we can avoid it, it would be best to not overwrite variables. So here, we could be calling it `siblingEl` instead of overwriting `sibling`. If the functionality needs to change later, refactoring might get confusing as the class name and the element are both called `sibling`.

src/resources/js/rsvp.js
```diff
@@ -1,37 +1,185 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	$( '.tribe-block__rsvp__ticket' ).on( 'click',
+		'.tribe-block__rsvp__status-button--going, .tribe-block__rsvp__status-button--not-going',
+		function( e ) {
 
+		var button    = $( this );
```
paulmskim: the spacing for this entire function should be tabbed over one more.

src/resources/js/rsvp.js
```diff
@@ -1,37 +1,185 @@
...
+			var input    = $( this ).parent().find( 'input[type="number"]' );
+			var increase = $( this ).hasClass( 'tribe-block__rsvp__number-input-button--plus' );
+
+			// stepUp or stepDown the input according to the button that was clicked
+			increase ? input[0].stepUp() : input[0].stepDown();
+
+			// Trigger the on Change for the input as it's not handled via stepUp() || stepDown()
+			input.trigger( 'change' );
+
+	} );
```
paulmskim: this should be tabbed over one more

bordoni:
I would like for us to tackle some JS changes later on. Since we want to get this tested.

src/resources/js/rsvp.js
```diff
@@ -1,37 +1,185 @@
...
+			function( response ) {
+				// Get the remaining number
+				var remaining = response.data.remaining;
+
+				// Update the remaining template part
+				ticket.find( '.tribe-block__rsvp__details .tribe-block__rsvp__availability' ).replaceWith( response.data.remaining_html );
+
+				ticket.find( '.tribe-block__rsvp__form' ).html( response.data.html );
+
+				if ( 1 > remaining ) {
```
paulmskim: could we make this `remaining === 0`? unless there is a reason why it needs to be `< 1`, this is a bit difficult to follow when written this way.

juanfra:
Thanks @bordoni @paulmskim, back to you guys

PR #267, opened 2018-08-24
Task/112809 pro plugin hook 2

nealfennimore:
Adds in use of [InnerBlocks](https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/components/inner-blocks/README.md) for rendering blocks from other plugins. This ends mitigating having to inject plugins via redux, and is a much cleaner solution IMO. I wish I found this sooner, as I did not see this in the docs until today. 

The blocks end up being locked in to the [parent block](https://wordpress.org/gutenberg/handbook/block-api/#parent-optional) via the [parent](https://github.com/moderntribe/events-gutenberg/compare/architecture-phase-two...task/112809-pro-plugin-hook-2?expand=1#diff-6a6c075882cbf951c1efb817a2ee77fbR31) key.

See [PluginBlockHooks](https://github.com/moderntribe/events-gutenberg/compare/architecture-phase-two...task/112809-pro-plugin-hook-2?expand=1#diff-df4277d6da221422467fff3e35b24db1) for new hook component and [how it's being used](https://github.com/moderntribe/events-gutenberg/compare/architecture-phase-two...task/112809-pro-plugin-hook-2?expand=1#diff-2cf7f4fd952737dc1fa4c26c94162867). It's been added to the `event-datetime` block in the `event` plugin [here](https://github.com/moderntribe/events-gutenberg/compare/architecture-phase-two...task/112809-pro-plugin-hook-2?expand=1#diff-a0139bfb4523b31ff51f7a1cd31a4657).

What ends up happening is that redux becomes the orchestrater, and allows other plugin blocks to be rendered automatically, when the plugin becomes available and the block is registered. We may want to eventually incorporate available blocks in each plugin reducer, as gutenberg will error if a plugin initializes, but does not have a block registered by the same name

https://www.useloom.com/share/d639d95a641c4ca5ad3f21ed9bfcde0c


plugins/common/src/modules/data/plugins/constants.js
```diff
@@ -0,0 +1,2 @@
+export const EVENTS_PLUGIN = 'events';
+export const EVENTS_PRO_PLUGIN = 'events_pro';
```
bordoni: `events-pro` is what we should use.

plugins/events/src/modules/elements/icons/tec.js
```diff
@@ -6,7 +6,7 @@ import React from 'react';
 /**
  * Internal dependencies
  */
-import TECIcon from 'icons/tec.svg';
+import TECIcon from '@moderntribe/events/src/resources/icons/tec.svg';
```
bordoni: Why not call it `pluginIcon`? Feel like a better variable name in general.

nealfennimore: It would need to be `PluginIcon` as it's importing the svg as a react component. I think it already adequately describes this. Maybe just change the export name in https://github.com/moderntribe/events-gutenberg/blob/master/src/modules/elements/icons/element.js?

PR #276, opened 2018-08-29
Update request mechanism by using `fetch`

mitogh:
Use `window.fetch` instead of `window.wp.apiRequest` to do request Promise based instead of using `jQuery` to handle an
`xhr` request.

This PR is isolated from the next step that is part of the ticket: https://central.tri.be/issues/112138

Next step (in a different PR) is to update the `withAPIData` and favor the store for data storage.


plugins/common/src/modules/store/middlewares/request/wp-request.js
```diff
@@ -2,16 +2,19 @@
...
+import { config } from '@moderntribe/common/utils/globals';
 import { types } from '@moderntribe/common/store/middlewares/request';
-import { getResponseHeaders } from '@moderntribe/common/utils/request';
-
-export const BASE = '/wp/v2';
 
 export default () => ( next ) => ( action ) => {
+	const REST = get( config(), 'REST', {} )
+	const { url = '', nonce = '' } = REST;
+	const BASE = `${ url }wp/v2`;
```
bordoni: We should fetch that `wp/v2` from PHP variable, since you can filter it.

plugins/common/src/modules/store/middlewares/request/wp-request.js
```diff
@@ -2,16 +2,19 @@
  * External dependencies
  */
 import { noop, get } from 'lodash';
+import 'whatwg-fetch';
```
bordoni: Why import this?

mitogh: It works as a `polyfill` providing support for browsers where `fetch` is not supported.

plugins/common/src/modules/store/middlewares/request/wp-request.js
mitogh: Perfect.

plugins/common/src/modules/store/middlewares/request/wp-request.js
bordoni: Thanks 💟

plugins/common/src/modules/store/middlewares/request/wp-request.js
```diff
@@ -38,18 +43,33 @@ export default () => ( next ) => ( action ) => {
...
+	actions.start( endpoint, params );
+
+	const headers = {
+		'Content-Type': 'application/json',
+		...get( params, 'headers', {} ),
+		'X-WP-Nonce': nonce,
+	};
 
-	return apiRequest( {
+	return fetch( endpoint, {
```
nealfennimore: @mitogh looks good! Do you think it would be good to migrate this over to async/await syntax?

mitogh: I would ❤️ to, and brilliant 💡 .

Thanks, _working on it_

mitogh:
:monkey-dance:  `async`/`await` has been added.

PR #277, opened 2018-08-29
Fix '.' as a command error for windows

nealfennimore:


nealfennimore:
Working on this for Nico - there's still some issues with dirs not being linked correctly on windows

nealfennimore:
Closing issue as still symlinking issues on windows. Nico is going to use zips for testing instead

bordoni:
I feel like our best bet is to always use the Zips or the Bot when it's eventually running.

PR #281, opened 2018-08-30
Ticket Purchase/Registration styles with attendee details.

juanfra:
https://central.tri.be/issues/110064
https://central.tri.be/issues/112199

The idea of this PR is to submit the different views of the registration/purchase "page".

I have implemented the shortcode `[gutti_tickets_purchase ticket="ID"]` to show these views. So, the idea is that if you just put that shortcode on a page, using a ticket ID, it'll display this page https://moderntribe.invisionapp.com/share/YCKH7MVWJTE#/screens/315523863 but for the ticket, you are passing as a parameter.

The idea is that these templates can later receive tickets and quantities, once we have the strategy on how the purchase flow will be.

NOTE: I'm also including JS and a few amendments to the ticket style from what we have on the sketch files, and a small js function to make the plus/minus buttons work. I'm also working on the availability check via AJAX, so it'll be great to review this PR as soon as possible so I can submit the other changes ASAP


juanfra:
@bordoni I'd super like you to check this one as soon as you can.

src/resources/js/tickets.js
```diff
@@ -0,0 +1,39 @@
+var tribe_tickets = {
```
bordoni: Let's try to use `tribe.tickets.block`

src/resources/js/tickets.js
```diff
@@ -0,0 +1,39 @@
...
+	num_attendees: 0,
+	event        : {}
+};
+
+(function( $, my ) {
+	'use strict';
+
+	var tribe_ticket = $( '.tribe-block__tickets' );
```
bordoni: When setting up jQuery variables we should always start with `$`.

src/resources/js/tickets.js
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	/**
+	 * Handle the number input + and - actions
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	$( '.tribe-block__tickets__item' )
```
bordoni: Let's have a `tribe.tickets.block.selectors` variable, with all the selectors we use here. 
E.g.:
```
( function ( $, obj ) {

	// Allows external usage of the selectors with out copying strings around.
	obj.selector = {
		container: .tribe-block__tickets,
		item: '.tribe-block__tickets__item'
	};

	var $containers = $( $obj.selector.container );

	// Always loop to allow multiple on the same page, even if the default is only one.
	$containers.each( function() {
		var $container = $( this );

		// Gives us a context to work with
		var $items = $container.find( obj.selector.item );
	} );
} )( jQuery, tribe.tickets.block )

```

src/views/blocks/tickets/registration/summary/description.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * This template renders the purchase summary description
+ *
+ * @version TBD
+ *
+ */
+?>
+<div class="tribe-block__tickets__registration__description">
+	<?php wpautop( esc_html_e( 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor.', 'events-gutenberg' ) ); ?>
```
bordoni: Forgotten?

src/views/blocks/tickets/registration/attendee/fields/checkbox.php
```diff
@@ -0,0 +1,59 @@
...
+}
+
+if ( ! $options ) {
+	return;
+}
+?>
+<div
+	class="tribe-field tribe-block__tickets__item__attendee__field__checkbox <?php echo $required ? 'tribe-tickets-meta-required' : ''; ?>"
+>
+	<label class="tribe-tickets-meta-label">
```
paulmskim: should use another element here instead, the `label` element isn't really labeling another HTML element. perhaps an appropriate heading tag?

src/views/blocks/tickets/registration/attendee/fields/radio.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * This template renders a Single Ticket content
+ * composed by Title and Description currently
+ *
+ * @version TBD
+ *
+ */
+$field = $this->get( 'field' );
```
paulmskim: spacing?

src/views/blocks/tickets/registration/attendee/fields/radio.php
```diff
@@ -0,0 +1,58 @@
...
+$attendee_id   = null;
+$value         = '';
+$is_restricted = false;
+$slug          = $field['slug'];
+?>
+<div
+	class="tribe-field tribe-block__tickets__item__attendee__field__radio <?php echo $required ? 'tribe-tickets-meta-required' : ''; ?>"
+>
+	<header class="tribe-tickets-meta-label">
+		<?php echo wp_kses_post( $field['label'] ); ?>
```
paulmskim: header is great here! I would suggest wrapping it a heading tag here if appropriate. same with the comment above.

src/views/blocks/tickets/registration/summary/title.php
```diff
@@ -0,0 +1,13 @@
+<?php
+/**
+ * This template renders the summary Title
+ *
+ * @version TBD
+ *
+ */
+?>
+<div class="tribe-block__tickets__registration__title">
+	<header><?php esc_html_e( 'Ticket Registration', 'events-gutenberg' ); ?></header>
```
paulmskim: should definitely use a heading tag within the `header` tag here.

src/views/blocks/tickets/registration/summary/description.php
juanfra: No, the thing here is that the view is used in a shortcode (in a page, post, wherever) and we don't have event_id to pull the info from. The idea is that you get to see the event info there: https://moderntribe.invisionapp.com/share/YCKH7MVWJTE#/screens/315523863

src/views/blocks/tickets/registration/attendee/fields/checkbox.php
juanfra: Could be, I grabbed the original code from what we now have in event tickets plus. The idea is that this label "serves" as the field title, so, perhaps a heading would be a good alternative.

src/resources/js/tickets.js
juanfra: Is it implemented somewhere else? Getting a syntax error if I use that

juanfra: Right, but where's tribe defined? I can see that we have something in ET right now, but they're all backend dependencies. I'm not sure what's that you want to do here.

src/resources/js/tickets.js
bordoni: @juanfra would like to confirm with you what do you think about using the style of JS above.
That is where I wanted our Frontend JS to go if possible. More aligned with the JS we've been delivering on the past few Feature Releases.

juanfra: @bordoni I'm fine with it. Not sure if it's super representative to have the selectors on one object once the file gets bigger and bigger. But it's better for maintainability. What I feel is that we need to improve the inline docs for all of our FE JS files. I mean, on the plugin side, there's different approaches, the ts, tt, and all that. I think now that we're starting to create some lines, we could improve documentation so, when this gets merged and we have to start fixing things on MRs, it's not too complicated to get on track of what and how things are done.

With regards to using the dollar sign before the jQuery variables, is that only to know that they're JQuery and not dom elements nor simple vars? I like it better without it, but I don't have any solid reason for that. It's like, "I like black shirts"... It's just because I like them better than red ones without reason.

juanfra:
@bordoni just added a check for the `tribe` obj and then the `tribe.tickets` until we actually define where this will be defined. The idea is to remove that later. But, if we don't have it there right now, none of the code will be testable, and I need to push some code and work on changes for the ticket availability checks.

juanfra:
Thanks @bordoni and @paulmskim for the reviews, great stuff ❤️

PR #284, opened 2018-08-31
Add AJAX availability check for the tickets FE

juanfra:
🎫 https://central.tri.be/issues/110067

The idea of this PR is to add an availability check for the tickets on the FE. Once we get the results, some HTML elements (input quantities, max, etc) will be updated.

This is a WIP as we don't have defined the strategy for the purchase flow.

At first, I created a per ticket check, but now I'm doing all at once sending an array with the IDs of the tickets we have.

@bordoni please let me know your thoughts, and see if this is what you had in mind.

### DIFF
--- a/src/Tribe/Blocks/Tickets.php
+++ b/src/Tribe/Blocks/Tickets.php
@@ -8,6 +8,8 @@ class Tribe__Events_Gutenberg__Blocks__Tickets
 extends Tribe__Events_Gutenberg__Blocks__Abstract {
 
 	public function hook() {
+		add_action( 'wp_ajax_ticket-availability-check', array( $this, 'ticket_availability' ) );
+		add_action( 'wp_ajax_nopriv_ticket-availability-check', array( $this, 'ticket_availability' ) );
 
 		add_shortcode( 'gutti_tickets_purchase', array( $this, 'render_shortcode_attendees' ) );
 	}
@@ -153,4 +155,40 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 		return $content;
 	}
 
+	/**
+	 * Check for ticket availability
+	 *
+	 * @since  TBD
+	 *
+	 * @param  array $tickets (IDs of tickets to check)
+	 *
+	 * @return int
+	 */
+	public function ticket_availability( $tickets = array() ) {
+
+		$response  = array( 'html' => '' );
+		$tickets   = tribe_get_request_var( 'tickets', array() );
+
+		// Bail if we receive no tickets
+		if ( empty( $tickets ) ) {
+			wp_send_json_error( $response );
+		}
+
+
+		// Parse the tickets and create the array for the response
+		foreach ( $tickets as $ticket_id ) {
+
+			$ticket    = Tribe__Tickets__Tickets::load_ticket_object( $ticket_id );
+			$available = $ticket->available();
+			$response['tickets'][ $ticket_id ]['available'] = $available;
+
+			// If there are no more available we will send the template part HTML to update the DOM
+			if ( 0 === $available ) {
+				$response['tickets'][ $ticket_id ]['unavailable_html'] = tribe( 'gutenberg.template' )->template( 'blocks/tickets/quantity-unavailable', $ticket, false );
+			}
+
+		}
+
+		wp_send_json_success( $response );
+	}
 }
--- a/src/Tribe/Blocks/Tickets.php
+++ b/src/Tribe/Blocks/Tickets.php
@@ -8,8 +8,8 @@ class Tribe__Events_Gutenberg__Blocks__Tickets
 extends Tribe__Events_Gutenberg__Blocks__Abstract {
 
 	public function hook() {
-		add_action( 'wp_ajax_ticket-availability-check', array( $this, 'ticket_availability' ) );
-		add_action( 'wp_ajax_nopriv_ticket-availability-check', array( $this, 'ticket_availability' ) );
+		add_action( 'wp_ajax_ticket_availability_check', array( $this, 'ticket_availability' ) );
+		add_action( 'wp_ajax_nopriv_ticket_availability_check', array( $this, 'ticket_availability' ) );
 
 		add_shortcode( 'gutti_tickets_purchase', array( $this, 'render_shortcode_attendees' ) );
 	}
--- a/src/Tribe/Blocks/Tickets.php
+++ b/src/Tribe/Blocks/Tickets.php
@@ -121,7 +121,7 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 	 *
 	 * @since  TBD
 	 *
-	 * @return void
+	 * @return string
 	 */
 	public function render_shortcode_attendees( $atts, $content = '' ) {
 
@@ -162,7 +162,7 @@ extends Tribe__Events_Gutenberg__Blocks__Abstract {
 	 *
 	 * @param  array $tickets (IDs of tickets to check)
 	 *
-	 * @return int
+	 * @return void
 	 */
 	public function ticket_availability( $tickets = array() ) {
 
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -17,11 +17,11 @@ tribe.tickets.block = {
 	'use strict';
 
 	obj.selector = {
-		container                    : '.tribe-block__tickets',
-		item                         : '.tribe-block__tickets__item',
-		itemQuantity                : '.tribe-block__tickets__item__quantity',
-		itemQuantityInput          : '.tribe-ticket-quantity',
-		itemExtraAvailable         : '.tribe-block__tickets__item__extra__available',
+		container                 : '.tribe-block__tickets',
+		item                      : '.tribe-block__tickets__item',
+		itemQuantity              : '.tribe-block__tickets__item__quantity',
+		itemQuantityInput         : '.tribe-ticket-quantity',
+		itemExtraAvailable        : '.tribe-block__tickets__item__extra__available',
 		itemExtraAvailableQuantity: '.tribe-block__tickets__item__extra__available_quantity'
 	};
 
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -19,10 +19,10 @@ tribe.tickets.block = {
 	obj.selector = {
 		container                    : '.tribe-block__tickets',
 		item                         : '.tribe-block__tickets__item',
-		item_quantity                : '.tribe-block__tickets__item__quantity',
-		item_quantity_input          : '.tribe-ticket-quantity',
-		item_extra_available         : '.tribe-block__tickets__item__extra__available',
-		item_extra_available_quantity: '.tribe-block__tickets__item__extra__available_quantity'
+		itemQuantity                : '.tribe-block__tickets__item__quantity',
+		itemQuantityInput          : '.tribe-ticket-quantity',
+		itemExtraAvailable         : '.tribe-block__tickets__item__extra__available',
+		itemExtraAvailableQuantity: '.tribe-block__tickets__item__extra__available_quantity'
 	};
 
 	var $tribe_ticket = $( obj.selector.container );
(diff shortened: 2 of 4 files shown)


src/Tribe/Blocks/Tickets.php
```diff
@@ -153,4 +155,40 @@ public function render_shortcode_attendees( $atts, $content = '' ) {
...
+	/**
+	 * Check for ticket availability
+	 *
+	 * @since  TBD
+	 *
+	 * @param  array $tickets (IDs of tickets to check)
+	 *
+	 * @return int
+	 */
+	public function ticket_availability( $tickets = array() ) {
```
bordoni: Prefix it with `ajax_` to make it easier to spot what is the usage for this.

src/Tribe/Blocks/Tickets.php
```diff
@@ -153,4 +155,40 @@ public function render_shortcode_attendees( $atts, $content = '' ) {
...
+			wp_send_json_error( $response );
+		}
+
+
+		// Parse the tickets and create the array for the response
+		foreach ( $tickets as $ticket_id ) {
+
+			$ticket    = Tribe__Tickets__Tickets::load_ticket_object( $ticket_id );
+			$available = $ticket->available();
+			$response['tickets'][ $ticket_id ]['available'] = $available;
```
bordoni: Should we always push the HTML? via importing the whole Quantity view?
That would prevent the logic from been done in two places.
But also return the `['available']` index for any other purposes; E.g.:

```
array(
    'available' => $ticket->available(),
    'html' => tribe( 'gutenberg.template' )->template( 'blocks/tickets/quantity', $ticket, false ),
)
```

So no matter what is the answer we replace the HTML on the correct place.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
+	}
+
+	/**
+	 * Check tickets availability
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	function check_availability() {
```
bordoni: I would make this and the other method available via the `obj.getTickets()` and `obj.checkAvailability()`

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
 	} );
 
+	/**
+	 * Get the tickets IDs
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	function get_tickets() {
```
bordoni: Naming convention should be CamelCase.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
+		};
+
+		$.post(
+			TribeTickets.ajaxurl,
+			params,
+			function( response ) {
+				var success = response.success;
+
+				// If we get a successful response
+				if ( success ) {
```
bordoni: Invert this logic to avoid indentation.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
+
+					// Get the tickets response with availability
+					var tickets = response.data.tickets;
```
bordoni: Lets make this whole logic here be a Method that we can pass in the `tickets`, would make the code more re-usable.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
+
+					Object.keys( tickets ).forEach( function( ticket_id ) {
+
+						var available = tickets[ ticket_id ].available;
+						var ticketEl = $( obj.selector.item + "[data-ticket-id='" + ticket_id + "']" );
```
bordoni: When selecting jQuery Objects start with the dollar sign e.g.: `$ticket`

src/resources/js/tickets.js
```diff
@@ -51,4 +51,81 @@ tribe.tickets.block = {
...
+
+							var unavailable_html = tickets[ ticket_id ].unavailable_html;
+							// Set the availability data attribute to false
+							ticketEl.attr( 'available', false );
+							// Remove classes for instock and purchasable
+							ticketEl.removeClass( 'instock' );
+							ticketEl.removeClass( 'purchasable' );
+
+							// Update HTML elements with the "Out of Stock" messages
+							ticketEl.find( '.tribe-block__tickets__item__extra__available' ).replaceWith( unavailable_html );
```
bordoni: Lets move these selectors into the `obj.selector`. All of the selectors should be available.

src/views/blocks/tickets/extra-available.php
```diff
@@ -11,5 +11,6 @@
 <div
 	class="tribe-block__tickets__item__extra__available"
 >
-	<?php echo sprintf( __( '%d available', 'events-gutenberg' ), $ticket->available() ); ?>
+	<span><?php echo esc_html( $ticket->available() ); ?></span>
```
bordoni: Lets make it a class?

src/views/blocks/tickets/item.php
```diff
@@ -14,12 +14,14 @@
...
 	'ticket' => $ticket,
-	'key' => $this->get( 'key' ),
+	'key'    => $this->get( 'key' ),
 );
 ?>
 <div
 	id="tribe-block-tickets-item-<?php echo esc_attr( $ticket->ID ); ?>"
 	class="<?php echo implode( ' ', get_post_class( $classes, $ticket->ID ) ); ?>"
+	data-ticket-id="<?php echo esc_attr( $ticket->ID ); ?>"
+	data-available="<?php echo ( 0 === $ticket->available() ) ? 'false' : 'true'; ?>"
```
bordoni: Should we make it a number?

src/Tribe/Blocks/Tickets.php
juanfra: Yes, thought about that but the idea is that we have to replace a couple chunks of HTML, not only in one place. 🤔

src/resources/js/tickets.js
juanfra: you don't like indentation at all 😂

src/views/blocks/tickets/item.php
juanfra: The thing is that we have `-1` for unlimited, so I preferred to keep this binary because we actually have the availability in the number input

src/resources/js/tickets.js
bordoni: Nopes. ❤️

src/Tribe/Blocks/Tickets.php
bordoni: Then we use the `available` field for the others, dunno might make it easier for that quantity part, because allows people to fully customize it

src/views/blocks/tickets/item.php
bordoni: Sounds good

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+				var tickets = response.data.tickets;
+
+				Object.keys( tickets ).forEach( function( ticket_id ) {
+
+					var available = tickets[ ticket_id ].available;
+					var ticketEl = $( obj.selector.item + "[data-ticket-id='" + ticket_id + "']" );
+
+					if ( 0 === available ) { // ticket is out of stock
+
+						var unavailable_html = tickets[ ticket_id ].unavailable_html;
```
paulmskim: you've got `snake_case` and `camelCase` here, should pick one.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+	 */
+	obj.checkAvailability = function() {
+
+		// We're checking availability for all the tickets at once
+		var params = {
+			action  : 'ticket-availability-check',
+			tickets : obj.getTickets(),
+		};
+
+		$.post(
```
paulmskim: do we have a timeout set for this? should we set a timeout? my main concern is as we have a `setInterval()` below of 15 seconds, if there is more than one call happening concurrently, how do we want to handle that (do we just allow it to play out)?

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+			function( response ) {
+				var success = response.success;
+
+				// Bail if we don't get a successful response
+				if ( ! success ) {
+					return;
+				}
+
+				// Get the tickets response with availability
+				var tickets = response.data.tickets;
```
bordoni: Lets move this into another method.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+
+				});
+
+			}
+		);
+
+	}
+
+	// Check tickets availability every 15 seconds
+	var check = setInterval( obj.checkAvailability, 15000 );
```
bordoni: Lets avoid the Interval, would be better to use a SetTimeout on the response of the AJAX request.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+						var unavailable_html = tickets[ ticket_id ].unavailable_html;
+						// Set the availability data attribute to false
+						ticketEl.attr( 'available', false );
+						// Remove classes for instock and purchasable
+						ticketEl.removeClass( 'instock' );
+						ticketEl.removeClass( 'purchasable' );
+
+						// Update HTML elements with the "Out of Stock" messages
+						ticketEl.find( '.tribe-block__tickets__item__extra__available' ).replaceWith( unavailable_html );
+						ticketEl.find( '.tribe-block__tickets__item__quantity' ).html( unavailable_html );
```
bordoni: Lets move these selectors into the `obj.selector`

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+
+						var unavailable_html = tickets[ ticket_id ].unavailable_html;
+						// Set the availability data attribute to false
+						ticketEl.attr( 'available', false );
+						// Remove classes for instock and purchasable
+						ticketEl.removeClass( 'instock' );
+						ticketEl.removeClass( 'purchasable' );
+
+						// Update HTML elements with the "Out of Stock" messages
+						ticketEl.find( '.tribe-block__tickets__item__extra__available' ).replaceWith( unavailable_html );
```
bordoni: Lets move these selectors into the `obj.selector`

src/resources/js/tickets.js
bordoni: Yeps. We should go with `camelCase`.

src/resources/js/tickets.js
```diff
@@ -51,4 +51,83 @@ tribe.tickets.block = {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	obj.checkAvailability = function() {
+
+		// We're checking availability for all the tickets at once
+		var params = {
+			action  : 'ticket-availability-check',
```
bordoni: We should use `_` for the divider since we will use in a Hook.

src/Tribe/Blocks/Tickets.php
juanfra: You mean for cases where it changes and they want to add a class or logic over there... something like setting a `.hot-selling` class if the availability is less than X?

juanfra: The thing here, I think, is that the **quantity template has the number input too**. 

So, for example, if somebody is trying to purchase a couple tickets, and reading the ticket description, takes more than 15 seconds and his selection will go back to zero.

juanfra:
Thanks @bordoni & @paulmskim!

I think the camelcasing and the use of obj for selectors and functions would be something we could add to the new JS specs.

bordoni:
Indeed @juanfra would you mind logging a ticket for me to handle that in the Sprint 2?

paulmskim:
Hey @juanfra thanks for changing the camel casing. Still the question on the JS API request and whether we're going to include a timeout, handle potential multiple requests, etc.

PR #286, opened 2018-08-31
Task/113412 derive datetime data

nealfennimore:
https://central.tri.be/issues/113412

This ticket sets up recurrence rules and exceptions blocks to utilize the start/end dates from the datetime block, so that the rule or exception defaults date/time to user's event.

- Add in redux saga
- Add in UI sagas for managing block visible and expand states
- Minor style fixes


nealfennimore:
Class names updated - needs another :eye:

PR #288, opened 2018-08-31
Feature/112919 block outside container

paulmskim:
Still a WIP, but almost there. A couple notes here:

- The button component could potentially move to the common elements.
- I was following similar classname structure that @nealfennimore was using for events pro, but could be changed to something else if we're all in agreement.

For ticket: https://central.tri.be/issues/112919

### DIFF
--- a/plugins/tickets/src/modules/elements/action-dashboard/element.js
+++ b/plugins/tickets/src/modules/elements/action-dashboard/element.js
@@ -1,0 +1,85 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import { Button } from '@moderntribe/tickets/elements'
+import './style.pcss';
+
+const ActionDashboard = ( {
+	actions,
+	cancelLabel,
+	className,
+	confirmLabel,
+	isCancelDisabled,
+	isConfirmDisabled,
+	onCancelClick,
+	onConfirmClick,
+	showCancel,
+	showConfirm,
+} ) => (
+	<section
+		className={ classnames(
+			'tribe-tickets__action-dashboard',
+			className,
+		) }
+	>
+		{ actions && actions.length && (
+			<div className="tribe-tickets__action-dashboard__group-left">
+				{ actions.map( ( action, index ) => (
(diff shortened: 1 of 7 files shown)


plugins/tickets/src/modules/elements/button/element.js
```diff
@@ -0,0 +1,37 @@
...
+		disabled={ isDisabled }
+		type={ type }
+		onClick={ onClick }
+	>
+		{ label }
+	</button>
+);
+
+Button.defaultProps = {
+	type: 'button',
```
mitogh: `isDisabled` might be a good one to set the default as well over here.

plugins/tickets/src/modules/elements/button/element.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const Button = ( {
```
bordoni: If it's a simple button maybe we should use one from Common or include this in common right?

plugins/tickets/src/modules/elements/button/element.js
paulmskim: default for `isDisabled` is `undefined` if not passed, which is falsy. I've left it out unless the default was anything other than falsy.

plugins/common/src/modules/elements/button/element.js
```diff
@@ -0,0 +1,38 @@
...
+		disabled={ isDisabled }
+		type={ type }
+		onClick={ onClick }
+	>
+		{ label }
+	</button>
+);
+
+Button.defaultProps = {
+	type: 'button',
```
mitogh: I will favor the usage of `noop` from lodash instead.

plugins/tickets/src/modules/elements/action-dashboard/element.js
```diff
@@ -0,0 +1,87 @@
...
+				) }
+			</div>
+		) }
+	</section>
+);
+
+ActionDashboard.defaultProps = {
+	showCancel: true,
+	showConfirm: true,
+	onCancelClick: () => {},
```
mitogh: Same as above; just for consistency reasons mainly

PR #295, opened 2018-09-04
Removal of icons, result from a merge operation

mitogh:
During the merge of `task/tickets-block` into this branch the icons files were added back into the `resources` directory however, now they are part of the module, this change of modules was introducing in this PR:

- https://github.com/moderntribe/events-gutenberg/pull/289


mitogh:
**Note:** No changelog entry was added as in theory the change shouldn't be part of this branch.

PR #297, opened 2018-09-05
Create container for the tickets block

mitogh:
- Create a container for the tickets block
- Create Template and Container components for tickets block
- Update `<Button>` test to use mock functions and test `clickHandlers`
- Update `<Button>` component to render the children and behave as a "native" HTML component
- Add `<Availability>` component used
- Add `<ActionButton>` component wrapper for `<Button>` but with styles for `<Icons>`


plugins/tickets/src/modules/blocks/tickets/Container.js
```diff
@@ -0,0 +1,39 @@
...
+import { Button } from '@moderntribe/common/elements';
+import { ActionButton } from '@moderntribe/tickets/elements';
+import {
+	Cog as CogIcon,
+	Tag as TagIcon,
+	User as UserIcon,
+} from '@moderntribe/common/src/modules/icons';
+
+const mapStateToProps = ( state, props ) => ( {
+	selected: props.isSelected,
```
paulmskim: I think we don't need this, as the prop `isSelected` passes through to the template component.

plugins/tickets/src/modules/blocks/tickets/Template.js
```diff
@@ -0,0 +1,77 @@
...
+		total,
+		footerActions,
+		footerConfirmLabel,
+	} = props;
+
+	const availability = selected && (
+		<Availability available={ available } total={ total } />
+	);
+
+	const actionDashboard = selected && (
```
paulmskim: re the note about `isSelected`, we can just use `isSelected` here.

plugins/tickets/src/modules/elements/availability/element.js
```diff
@@ -0,0 +1,51 @@
...
+import { __ } from '@wordpress/i18n';
+import { identity } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import { numericLabel } from '@moderntribe/common/utils/string';
+import './style.pcss';
+
+const TicketAvailability = ( { available, total, separator, ...labels } ) => {
```
paulmskim: If our element name in the folder structure is `availability`, then the component name should follow this as well. So the folder either becomes `elements/ticket-availability/` or the component name becomes `Availability`.

plugins/tickets/src/modules/elements/availability/element.js
```diff
@@ -0,0 +1,51 @@
...
+			count: available, singular: labels.availableSingular, plural: labels.availablePlural,
+		} ),
+		numericLabel( {
+			count: total, singular: labels.totalSingular, plural: labels.totalPlural,
+		} ),
+	].filter( identity )
+		.join( separator );
+
+	return (
+		<div className="tribe-editor__availability">{ items }</div>
```
paulmskim: Same note here on the component name, if the component name becomes `TicketAvailability` this should also become `tribe-editor__ticket-availability`.

plugins/tickets/src/modules/elements/availability/element.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+TicketAvailability.defaultProps = {
+	available: 0,
+	total: 0,
+	availableSingular: __( '%d ticket available', 'events-gutenberg' ),
+	availablePlural: __( '%d tickets available', 'events-gutenberg' ),
+	totalSingular: __( '%d total ticket', 'events-gutenberg' ),
+	totalPlural: __( '%d total tickets', 'events-gutenberg' ),
+	separator: ' | ',
```
paulmskim: the separator should only be a separator, no spaces.

plugins/tickets/src/modules/elements/availability/element.js
```diff
@@ -0,0 +1,51 @@
...
+import { identity } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import { numericLabel } from '@moderntribe/common/utils/string';
+import './style.pcss';
+
+const TicketAvailability = ( { available, total, separator, ...labels } ) => {
+	const items = [
```
paulmskim: We should be wrapping each piece in a `span` here. Currently, you would output something like:

```js
<div className="tribe-editor__availability">48 tickets available | 166 tickets total</div>
```

A couple things that are going to be difficult here:
- the numbers need to be bold and larger font size than the text, that'll need a separate style.
- the spacing between the separator and the numeric labels may not be one space width, and using spaces to make up that spacing is not good practice.

I would, for starters, recommend wrapping every piece of text that has a different style or spacing in a `span` so you can control them much more easily.

plugins/tickets/src/modules/elements/availability/element.js
```diff
@@ -0,0 +1,51 @@
...
+		numericLabel( {
+			count: available, singular: labels.availableSingular, plural: labels.availablePlural,
+		} ),
+		numericLabel( {
+			count: total, singular: labels.totalSingular, plural: labels.totalPlural,
+		} ),
+	].filter( identity )
+		.join( separator );
```
paulmskim: this separator should be wrapped in a `span`

plugins/common/src/modules/elements/button/style.pcss
```diff
@@ -0,0 +1,7 @@
+.tribe-editor__btn--label {
```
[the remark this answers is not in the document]
mitogh: The style is not added to the button is present to be added manually, however as is a common use case the base style is part of the `common.css` instead.

plugins/common/src/modules/utils/__tests__/string.test.js
```diff
@@ -5,7 +5,8 @@ import {
 	isTruthy,
 	isFalsy,
 	replaceWithObject,
-} from '@moderntribe/events/editor/utils/string';
+	numericLabel,
+} from '../string';
```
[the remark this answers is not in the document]
mitogh: Perfect.

plugins/tickets/src/modules/elements/action-button/__tests__/element.test.js
```diff
@@ -0,0 +1,24 @@
...
+ * External dependencies
+ */
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+/**
+ * Internal dependencies
+ */
+import { ActionButton } from '@moderntribe/tickets/elements';
+import { Button } from '@moderntribe/common/elements';
```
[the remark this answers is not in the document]
mitogh: Yes is used here: 

- https://github.com/moderntribe/events-gutenberg/blob/d46a67c3a67110cb3fd6981660711217d179e2cb/plugins/tickets/src/modules/elements/action-button/__tests__/element.test.js#L21

plugins/tickets/src/modules/blocks/tickets/Container.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * External dependencies
+ */
+import { connect } from 'react-redux';
+import { compose, bindActionCreators } from 'redux';
```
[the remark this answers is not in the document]
mitogh: No for now, just boilerplate but I agree we should remove it for now.

plugins/tickets/src/modules/blocks/tickets/Container.js
mitogh: Yeah however here is used as a proxy by using `selected` instead of `isSelected`

plugins/tickets/src/modules/elements/availability/element.js
mitogh: Yep my bad will update to match file name.

plugins/tickets/src/modules/elements/availability/element.js
mitogh: Answer above so no need to change here.

plugins/tickets/src/modules/elements/action-button/element.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+import { Button } from '@moderntribe/common/src/modules/elements';
+import './style.pcss';
+
+const ActionButton = ( { children, ...props } ) => (
+	<Button
+		className={ classNames( 'tribe-editor__btn--label', 'tribe-editor__tickets-btn--action' ) }
+		{ ...props }
+	>
+		{ children }
```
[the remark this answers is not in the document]
mitogh: Yeah that was my initial though on this one however by inverting the control of how the inside looks you can have the icon on the right side as well.

I think main difference and benefit of having a wrapper around `<Button>` is the styling of the button as has: 

1. A style that removes all button look like
1. A style that matches footer design and separates icon from label

mitogh:
Changes has been added.

plugins/tickets/src/modules/elements/action-button/element.js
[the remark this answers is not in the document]
paulmskim: If you want the icon on the right side vs the left side, then we should add a prop for this, say something like `iconPosition` with enums `'left'` or `'right'` as possible values. Also, as icons are not meaningful content for screenreaders and other accessibility devices, we can flip around the icon and label using CSS.

Classes can be provided to the `Button` component directly and override styles, so I don't think this should be our main reason for providing a wrapper. If we're changing the structure, I find that to be a more meaningful reason for a new element.

plugins/tickets/src/modules/blocks/tickets/Container.js
paulmskim: Right, I understand what's happening here. What I'm saying is that even without the proxy, this property is automatically provided to the template component.

If you have a look at https://github.com/reduxjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options under `mergeProps` you'll find that if you omit `mergeProps` argument to `connect()`, it'll pass `Object.assign({}, ownProps, stateProps, dispatchProps)` by default. `isSelected` is part of `ownProps`, so it's provided to the template by default.

plugins/common/src/modules/elements/button/style.pcss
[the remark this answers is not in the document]
paulmskim: got it, I saw that as I was going through the files.

plugins/tickets/src/modules/elements/action-button/__tests__/element.test.js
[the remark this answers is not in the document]
mitogh: In this tests we don't simulate any click on `Button` instead `Button` is being used as selector to find the element and ensure that element has the correct class applied.

plugins/tickets/src/modules/elements/action-button/element.js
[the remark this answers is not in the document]
mitogh: Great I think will update `ActionButton` to have a defined structure instead as I think `<Button>` should remain closest to the native `<button>` as possible an any sort of variation of this should be a wrapper instead, to provide higher flexibility.

plugins/tickets/src/modules/elements/action-button/__tests__/element.test.js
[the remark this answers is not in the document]
paulmskim: You're right, I was thinking of another test.

plugins/common/src/modules/elements/button/element.js
```diff
@@ -3,23 +3,28 @@
...
 } ) => (
 	<button
-		className={ classnames( 'tribe-editor__button', className ) }
+		className={ classNames( 'tribe-editor__button', className ) }
 		disabled={ isDisabled }
 		type={ type }
 		onClick={ onClick }
 	>
-		{ label }
+		{ children }
```
paulmskim: one thing to note about this is that this change will be breaking for any component that uses the button element. this is something to be aware of when merging.

mitogh: I think I did an update originally but will update again, thanks.

plugins/tickets/src/modules/elements/numeric-label/element.js
```diff
@@ -0,0 +1,60 @@
...
+const NumericLabel = ( { count, singular, plural, fallback, className } ) => {
+	if ( count <= 0 ) {
+		return fallback;
+	}
+	const regex = /%d/gi;
+	const targetStr = count === 1 ? singular : plural;
+	const [ before, after ] = targetStr.split( '%d' );
+	return (
+		<span className={ classNames( 'tribe-editor__label--numeric', className ) }>
+			{ before && <span className="tribe-editor__label--before">{ before }</span> }
```
paulmskim: this element is much better with the spans. could we do a couple things here:
- make the classes `tribe-editor__numeric-label--` rather than just `__label`, if we wanted another `label` component this will make things clear.
- put a space after the `{ before }` but before the `</span>` and put a space before the `{ after }` but after the `<span>`, so it looks like `...>{ before } </span>` and `> { after }</span>`. this preserves the proper spacing for accessibility.

plugins/tickets/src/modules/elements/action-button/element.js
```diff
@@ -0,0 +1,46 @@
...
+		'tribe-editor__tickets-btn--action',
+		`tribe-editor__tickets-btn--action-${ position }`,
+	);
+
+	return (
+		<Button
+			className={ containerClass }
+			{ ...props }
+		>
+			{ icon }{ children }
```
paulmskim: let's wrap this `{ children }` in a span. In the case `children` is only text, we want to be able to apply styles specifically to the text and not the entire button.

plugins/tickets/src/modules/elements/numeric-label/element.js
paulmskim: sorry disregard the 2nd point, I wasn't thinking. the split string should already have the spaces.

plugins/tickets/src/modules/elements/action-button/element.js
mitogh: Good idea.

mitogh:
Changes has been addressed, let me know if you have additional questions.

mitogh:
Merging this one @paulmskim as latest changes has been addressed and we need to merge this one to unblock the other PRs. 

Thanks for the review.

plugins/tickets/src/modules/elements/numeric-label/element.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param {string} singular The label for the singular case
+ * @param {string} plural The label for the plural case
+ * @param {*} fallback The value to be returned if count is zero or negative
+ * @returns {*} return fallback if count is zero or negative otherwise singular or plural
+ */
+const NumericLabel = ( { count, singular, plural, fallback, className } ) => {
+	if ( count <= 0 ) {
+		return fallback;
+	}
+	const regex = /%d/gi;
```
paulmskim: @mitogh if this isn't used, we should remove this. if you could get this in your next pr that'd be great!

mitogh: removing this one on the next PR

PR #299, opened 2018-09-06
Fix/action dashboard render

paulmskim:
Fixes a couple style and render issues around action dashboard.

### DIFF
--- a/plugins/tickets/src/modules/elements/action-dashboard/element.js
+++ b/plugins/tickets/src/modules/elements/action-dashboard/element.js
@@ -30,38 +30,46 @@ const ActionDashboard = ( {
 			className,
 		) }
 	>
-		{ actions && actions.length && (
-			<div className="tribe-editor__action-dashboard__group-left">
-				{ actions.map( ( action, index ) => (
-					<span
-						key={ `action-${index}`}
-						className="tribe-editor__action-dashboard__action-wrapper"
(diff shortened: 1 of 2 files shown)


plugins/tickets/src/modules/elements/action-dashboard/element.js
```diff
@@ -30,38 +30,46 @@ const ActionDashboard = ( {
...
-					<Button
-						className="tribe-editor__action-dashboard__confirm-button"
-						isDisabled={ isConfirmDisabled }
-						label={ confirmLabel }
-						onClick={ onConfirmClick }
-					/>
-				) }
-			</div>
-		) }
+		{
```
mitogh: I think having this type of structure is helpful when the logic of the component is simpler and by type of structure I mean the explicit return, however I think having this function wrapped in braces will help to improve clarity on this one due the size of the component.

```js
() => {
	const actions = actions && actions.length 
		? ()
		: null;
	const groupRight = ...

	return (
		<div>
			{ actions }
			{ groupRight }
		</div>
	);
}
```

bordoni: I would love to try to simplify it like that too.

paulmskim: done

PR #300, opened 2018-09-06
Thunks to sagas price block

nealfennimore:
Converts thunks in price block to sagas

https://central.tri.be/issues/113725


nealfennimore:
@mitogh is that what you mean?

mitogh:
I mean the file name `sagas.spec.js` as most of the test at this point are named like `sagas.test.js` just curios on the reason behind that.

Thakns

nealfennimore:
Ah just a habit of mine - I'll change it

PR #301, opened 2018-09-06
Do a shallow comparision between elements

mitogh:
For elements like array or objects when values are the same as they are
objects the references to those objects are different so doing a strict
comparission returns false eve if the values of each is the same

Ticket: https://central.tri.be/issues/112578


mitogh:
Would you mind take another look @nealfennimore 

As tests were failing on the other one.

Thanks

PR #306, opened 2018-09-10
Add <CapacityTable> component for Tickets

mitogh:


plugins/tickets/src/modules/elements/capacity-table/row.js
```diff
@@ -0,0 +1,31 @@
...
+			{ items && <span className="tribe-editor__capacity-row__items">{ items }</span> }
+		</span>
+		<span className="tribe-editor__capacity-row--right">{ right }</span>
+	</div>
+);
+
+CapacityRow.propTypes = {
+	label: PropTypes.string,
+	items: PropTypes.string,
+};
```
paulmskim: missing proptypes for `right`?

plugins/tickets/src/modules/elements/capacity-table/style.pcss
```diff
@@ -0,0 +1,40 @@
...
+	display: flex;
+	justify-content: space-between;
+
+	&:nth-last-of-type(2) {
+		border-bottom: 1px solid #E1E3E6;
+	}
+}
+
+.tribe-editor__capacity-row--left {
+	flex-grow: 1;
```
paulmskim: should define `flex` property here, not just `flex-grow`.

plugins/tickets/src/modules/elements/capacity-table/style.pcss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.tribe-editor__capacity-row--right {
+	flex-basis: 10%;
```
paulmskim: same as above, define `flex` not `flex-basis`.

readme.md
```diff
@@ -105,6 +105,7 @@ Finally, run `npm run bootstrap` from the root to link the plugin up.
 * Feature - Create `Available` component
 * Feature - Add TribePropTypes functions to check custom time format
 * Feature - Add time to seconds and ms and vice versa utility
+* Feature - Create `<HeaderImage>` component to upload / select an image from the media library
```
paulmskim: need changelog entry for capacity table

plugins/tickets/src/modules/elements/capacity-table/utils.js
```diff
@@ -0,0 +1,6 @@
+import identity from 'lodash/identity';
+
+export const toLabel = ( items = [] ) => {
```
paulmskim: maybe add a docblock or a description to describe what's happening here.

plugins/tickets/src/modules/elements/capacity-table/style.pcss
```diff
@@ -14,19 +14,29 @@
...
+	flex: 1;
 }
 
 .tribe-editor__capacity-row--right {
-	flex-basis: 10%;
+	flex: none;
+	width: 10%;
 	text-align: right;
+
+	input {
```
paulmskim: Curious why we have an input style here. this almost expects that an `input` node is to be passed into `right`. But then this isn't really an element style then, and the styles know more than the element should. Could we maybe move this style to the parent component that uses this element instead?

mitogh:
Thanks changes has been updated.

PR #310, opened 2018-09-13
Add <Ticket> block

mitogh:
Create a child block available to insert a new ticket inside of the tickets list


mitogh:
*Note* Most of the items inside of the ticket item are just used as a placeholder  for now in order to render all the available options.

plugins/common/src/modules/utils/number.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Calculate the percentage of two numbers
+ *
+ * @param {number} value Initial value from where to take the percentage
+ * @param {number} total Total value to get the percentage relative to this value
+ * @returns {number} total percentage value
+ */
+export const percentage = ( value = 0, total = 0 ) => {
```
paulmskim: curious here, wouldn't it be OK here to allow for negative percentages and percentages above 100? this would make sense to me as both are quite possible and real.

plugins/common/src/modules/utils/number.js
```diff
@@ -0,0 +1,21 @@
...
+	if ( total <= 0 ) {
+		return 0;
+	}
+
+	const result = Math.floor( ( value / total ) * 100 );
```
paulmskim: there's a number utility `toFixed()` that can round to certain number of decimal places, it might allow for more flexibility for the user if they want a specific format.

plugins/common/src/modules/utils/proptypes.js
```diff
@@ -58,6 +58,15 @@ export const timeFormat = ( props, propName, componentName ) => {
...
+	if ( null !== props[ propName ] ) {
+		return new Error(
+			`Invalid prop: \`${propName}\` supplied to \`${ componentName }\`, expect null.`
+		);
+	}
+}
+
 export default {
 	timeFormat: createChainableValidator( timeFormat ),
+	nullType,
```
paulmskim: let's chain this to allow for `.isRequired` here by wrapping it in `createChainableValidator()`. we may not need it now, but just good practice.

plugins/tickets/src/modules/blocks/ticket/Container.js
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Template from './Template';
+
+import { withSaveData, withStore } from '@moderntribe/common/src/modules/hoc';
+
+const mapStateToProps = ( state, props ) => ( {
```
paulmskim: let's use the original function signature and use `ownProps` rather than `props`

plugins/tickets/src/modules/blocks/ticket/Template.js
```diff
@@ -0,0 +1,88 @@
...
+		unlimited,
+		expires,
+		available,
+		sold,
+	} = props;
+
+	const priceLabel = [ currencySign, price ];
+
+	return (
+		<div className="tribe-editor__ticket">
```
paulmskim: `article` tag is a good use here.

plugins/tickets/src/modules/blocks/ticket/Template.js
```diff
@@ -0,0 +1,88 @@
...
+					<button className="tribe-editor__btn--label"><Pencil /></button>
+					<button className="tribe-editor__btn--label"><Clipboard /></button>
+				</div>
+				{ description && <div className="tribe-editor__ticket-description">{ description }</div> }
+			</div>
+			<div className="tribe-editor__ticket-price">
+				{ 'suffix' === currencyPosition ? [ ...priceLabel ].reverse() : priceLabel }
+			</div>
+			<div className="tribe-editor__ticket-quantity">
+				{ `${available} of ${ sold } sold` }
```
paulmskim: let's wrap this in a span to allow for styling.

plugins/tickets/src/modules/blocks/ticket/Template.js
```diff
@@ -0,0 +1,88 @@
...
+				<QuantityBar sold={ 40 } total={ 156 } />
+				<span className="tribe-editor__quantity-label">7 of 10 sold <Alert /></span>
+				<QuantityBar sold={ 40 } shared={ 120 } total={ 156 } />
+				<QuantityBar sold={ 40 } capacity={ 70 } shared={ 120 } total={ 156 } />
+				<span className="tribe-editor__quantity--unlimited">unlimited</span>
+			</div>
+		</div>
+	);
```
paulmskim: this HTML here is really well written, awesome stuff!

plugins/tickets/src/modules/blocks/ticket/style.pcss
```diff
@@ -0,0 +1,60 @@
...
+	justify-content: space-between;
+	align-items: center;
+	/*
+	this negative margin is to fit the blocks to the bounds of the pseudo element :before of
+	the drag block
+	 */
+	margin: -14px;
+}
+
+.tribe-editor__ticket-icon {
```
paulmskim: need `flex` property for `__ticket-icon`, `__ticket-content`, and `__ticket-price`.

plugins/tickets/src/modules/blocks/ticket/style.pcss
```diff
@@ -0,0 +1,60 @@
...
+	display: flex;
+	align-items: center;
+	padding: 0 7px;
+
+	svg {
+		flex: none;
+	}
+}
+
+.tribe-editor__ticket-description {
```
paulmskim: need `flex` property here

plugins/tickets/src/modules/elements/quantity-bar/bar.js
```diff
@@ -0,0 +1,48 @@
...
+ * External dependencies
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import TribeTypes from '@moderntribe/common/utils/proptypes';
```
paulmskim: `TribePropTypes` is exported from `@moderntribe/common/utils`, you can just add it to the line below:
`import { number, TribePropTypes } from '@moderntribe/common/utils'`

plugins/tickets/src/modules/elements/quantity-bar/bar.js
```diff
@@ -0,0 +1,48 @@
...
+import { number } from '@moderntribe/common/utils/';
+
+const Bar = ( { className, value, total } ) => {
+
+	if ( value === 0 || total === 0 ) {
+		return null;
+	}
+
+	const percentage = number.percentage( value, total );
```
paulmskim: re the comment on the precentage util, I think this would be the right place to check for whether `total > 0` and `value > total`.

plugins/tickets/src/modules/elements/quantity-bar/bar.js
```diff
@@ -0,0 +1,48 @@
...
+	return (
+		<span className={ classNames( className ) } style={ style }></span>
+	);
+}
+
+Bar.propTypes = {
+	className: PropTypes.oneOfType( [
+		PropTypes.string,
+		PropTypes.arrayOf( PropTypes.string ),
+		TribeTypes.nullType,
```
paulmskim: curious why we need to check for `null` here. if you pass `null` into `classNames`, it shouldn't return anything anyway.

plugins/tickets/src/modules/elements/quantity-bar/style.pcss
```diff
@@ -0,0 +1,56 @@
...
+	background-color: #009FD4;
+	z-index: 2;
+}
+
+.tribe-editor__quantity-bar--shared {
+	background-color: #C9EBF7;
+	z-index: 1;
+}
+
+.tribe-editor__quantity-label {
```
paulmskim: this and `tribe-editor__quantity--unlimited` below should go in `plugins/tickets/src/modules/blocks/ticket/style.pcss`, not in the element styles.

plugins/tickets/src/modules/elements/quantity-bar/bar.js
mitogh: Basically to set the null inside of `defaultProps` instead of empty string.

plugins/common/src/modules/utils/number.js
mitogh: Right, I think the function should only be in charge of generate the percentage regardless of the boundaries and any specific boundary should be applied where is used to allow for more flexibility.

plugins/common/src/modules/utils/number.js
mitogh: Will create a param for this and same as above can be configure where is used instead.

plugins/common/src/modules/utils/number.js
mitogh: However you can have negative percentages if your total is `0` the percentage is gonna be zero as a division by zero is not defined.

paulmskim: definitely have a boundary for total = 0.

plugins/tickets/src/modules/blocks/ticket/Template.js
```diff
@@ -27,21 +36,29 @@ const Ticket = ( props ) => {
 	return (
 		<div className="tribe-editor__ticket">
```
paulmskim: article tag would probably be a good use here.

plugins/tickets/src/modules/elements/quantity-bar/bar.js
paulmskim: Still wondering why we need the `null` though. If we don't set a `defaultProps` for `className`, if we don't pass in a `className`, it should be `undefined`, which shouldn't display a class name anyway, no?

plugins/common/src/modules/utils/number.js
```diff
@@ -0,0 +1,16 @@
...
+ * @returns {number} total percentage value
+ */
+export const percentage = ( value = 0, total = 0 ) => {
+	if ( total === 0 ) {
+		return 0;
+	}
+
+	const result = Number.parseFloat( ( value / total ) * 100 );
+
+	return isNaN( result ) ? 0 : result;
```
paulmskim: last thing here, and the same above where `if ( total === 0 )`, if `result` is `NaN`, we shouldn't be returning 0 as that doesn't make any sense. We should probably return `NaN` or `false` to separate a bad result from a result that is actually 0.

paulmskim:
Just the one above and this left:
https://github.com/moderntribe/events-gutenberg/pull/310#discussion_r217507826

plugins/common/src/modules/utils/number.js
mitogh: I think throwing an error would be better I think as the function can just return `numbers` all the time instead and we can `try/catch` instead.

plugins/tickets/src/modules/elements/quantity-bar/bar.js
mitogh: That's right. I mean where this one is defined in `defaultProps` is mainly used as `null`

plugins/common/src/modules/utils/number.js
paulmskim: perfect, this works too.

PR #315, opened 2018-09-14
Feature/112953 rsvp store definition

paulmskim:
For ticket: https://central.tri.be/issues/112953


package.json
```diff
@@ -64,6 +64,7 @@
 		"redux-saga": "^0.16.0",
 		"redux-thunk": "^2.3.0",
 		"reselect": "^3.0.1",
+		"uniqid": "^5.0.3",
 		"url-loader": "^0.6.2",
 		"uuid": "^3.3.2",
```
bordoni: Are we going to remove `uuid`?

paulmskim: yeah good call, was going to but forgot about it.

plugins/tickets/src/modules/data/blocks/rsvp/thunks.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Internal dependencies
+ */
+import { actions } from '@moderntribe/tickets/data/blocks/rsvp';
+
+export const setRSVPHeader = ( data ) => ( dispatch ) => {
```
mitogh: We should start adding only sagas instead as is already available.

plugins/tickets/src/modules/blocks/rsvp/rsvp-advanced-options/template.js
```diff
@@ -3,6 +3,7 @@
  */
 import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
+import uniqid from 'uniqid';
```
mitogh: Wonder if the one provided by `lodash` is not enough for this one

plugins/common/src/modules/utils/moment.js
```diff
@@ -216,6 +216,10 @@ export const toTime = ( date, format = dateUtil.FORMATS.WP.time ) => {
 	return date.format( toFormat( format ) );
 }
 
+export const toTime24Hr = ( date, format = dateUtil.FORMATS.WP.time24Hr ) => {
```
mitogh: Wonder if we can just return directly without the need of the explicit `return` here instead.

plugins/tickets/src/modules/data/blocks/rsvp/actions.js
```diff
@@ -0,0 +1,68 @@
...
+} );
+
+export const setRSVPCapacity = ( capacity ) => ( {
+	type: types.SET_RSVP_CAPACITY,
+	payload: {
+		capacity,
+	},
+} );
+
+export const setRSVPEnableNotGoing = ( enableNotGoing ) => ( {
```
mitogh: Does this one also handles the `disable` case?

paulmskim: so this is actually a field called `enableNotGoing`. this action just sets the value, either `true` or `false`.

mitogh: In that case I think something like `setRSVPNotGoing` makes more sense instead

plugins/tickets/src/modules/blocks/rsvp/rsvp-advanced-options/template.js
paulmskim: i suppose, though lodash's uniqueid module only provides incremental numbers. I'd be worried that it won't be unique enough. also uniqid is lighter than lodash's uniqueid.

paulmskim:
@mitogh fixed up your suggestions, would like another pass before merging.

PR #317, opened 2018-09-16
Feature/112920 rsvp container

paulmskim:
For ticket: https://central.tri.be/issues/112920

Depends on https://github.com/moderntribe/events-gutenberg/pull/315 and requires it to be merged first. I suggest merging the other PR first before reviewing this as the diffs will show many that are the same.


mitogh:
Would you mind merging back `feature/event-ticket-blocks` into this one to have a cleaner diff.

Thanks @paulmskim

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/attendees-action-button/template.js
```diff
@@ -0,0 +1,33 @@
...
+import { User as UserIcon } from '@moderntribe/common/src/modules/icons';
+
+const AttendeesActionButton = ( {
+	onClick,
+} ) => (
+	<ActionButton
+		className="tribe-editor__rsvp__action-button tribe-editor__rsvp__action-button--attendees"
+		icon={ <UserIcon /> }
+	>
+		Attendees
```
mitogh: I think this one might be a good one to set as a prop as well and the default into `defaultProps`

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/settings-action-button/template.js
```diff
@@ -0,0 +1,33 @@
...
+import { Cog as CogIcon } from '@moderntribe/common/src/modules/icons';
+
+const SettingsActionButton = ( {
+	onClick,
+} ) => (
+	<ActionButton
+		className="tribe-editor__rsvp__action-button tribe-editor__rsvp__action-button--settings"
+		icon={ <CogIcon /> }
+	>
+		Settings
```
mitogh: This one ideally any string should be set as a prop in order to be translatable later.

plugins/tickets/src/modules/elements/container-panel/element.js
```diff
@@ -0,0 +1,46 @@
...
+	rightSectionHeader,
+} ) => (
+	<div
+		className={ classNames(
+			'tribe-editor__container-panel',
+			className,
+		) }
+	>
+		<section className="tribe-editor__container-panel__left-section">
+			{ leftSection }
```
mitogh: I think these items should be named: 

- `icon`
- `content`
- `body`

As discussed in the call.

plugins/tickets/src/modules/elements/container-panel/__tests__/element.test.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import renderer from 'react-test-renderer';
```
mitogh: No need to import this one anymore as is available globally via `jest.setup`

plugins/common/src/modules/elements/checkbox/__tests__/element.test.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import renderer from 'react-test-renderer';
```
mitogh: No need to import this one as is available globally

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
```diff
@@ -0,0 +1,92 @@
...
+ * Internal dependencies
+ */
+import RSVPAdvancedOptions from '@moderntribe/tickets/blocks/rsvp/rsvp-advanced-options/container';
+import { Checkbox } from '@moderntribe/common/elements';
+import './style.pcss';
+
+const RSVPContainerContentLabels = () => (
+	<div className="tribe-editor__rsvp-container-content__labels">
+		<span className="tribe-editor__rsvp-container-content__capacity-label">
+			{ __( 'RSVP Capacity', 'events-gutenberg' ) }
```
mitogh: I think those strings might be a good case for props as well.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
```diff
@@ -0,0 +1,92 @@
...
+			id={ capacityId }
+			min="0"
+			type="number"
+			value={ capacity }
+		/>
+		<Checkbox
+			checked={ notGoingResponses }
+			className="tribe-editor__rsvp-container-content__not-going-responses"
+			id={ notGoingId }
+			label={ __( 'Enable Not Going responses', 'events-gutenberg' ) }
```
mitogh: Same as above, string as prop.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
```diff
@@ -0,0 +1,92 @@
...
+);
+
+RSVPContainerContentOptions.propTypes = {
+	capacity: PropTypes.string,
+	capacityId: PropTypes.string.isRequired,
+	notGoingId: PropTypes.string.isRequired,
+	notGoingResponses: PropTypes.bool,
+};
+
+class RSVPContainerContent extends PureComponent {
```
mitogh: Wonder on this one, specifically because of the non stateless component.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-header/template.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Internal dependencies
+ */
+import { NumericLabel } from '@moderntribe/tickets/elements';
+import './style.pcss';
+
+const getCapacityLabel = ( capacity ) => {
+	const singular = __( '%d available', 'events-gutenberg' );
+	const plural = singular;
+	const fallback = null;
```
mitogh: I think this one is the default value so it might not be required to be set

plugins/tickets/src/modules/blocks/rsvp/template.js
```diff
@@ -0,0 +1,63 @@
...
+} from './rsvp-action-buttons';
+import { ActionDashboard } from '@moderntribe/tickets/elements';
+import './style.pcss';
+
+const RSVP = ( {
+	created,
+	isSelected,
+} ) => {
+	const getActionDashboard = () => {
+		const actions = [
```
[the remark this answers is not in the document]
paulmskim: I made changes to this in a later PR (pull it out into its own component), so I would rather make changes there.

plugins/tickets/src/modules/blocks/rsvp/template.js
```diff
@@ -0,0 +1,63 @@
...
+			<SettingsActionButton />,
+			<AttendeesActionButton />,
+		];
+		const confirmLabel = created
+			? __( 'Update RSVP', 'events-gutenberg' )
```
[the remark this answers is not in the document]
paulmskim: These strings only depend on one thing: whether the RSVP was created or not. If not, We aren't allowing the user to pass in a string or having it be editable, they are set as either of these 2.

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/attendees-action-button/template.js
paulmskim: Not sure what you mean here? The `Attendees` string is what is displayed, and this is the same as how you have it set in your action buttons.

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/settings-action-button/template.js
paulmskim: I realized this in a later PR, this change was made along with `Attendees`.

plugins/tickets/src/modules/elements/container-panel/element.js
paulmskim: Will do in another ticket, as we discussed.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
paulmskim: this is needed, because of the `this.capacityId` and `this.notGoingId` defined below.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
paulmskim: Where do you see this and the below `Enable Not Going responses` string to be changing state? I could see the argument for supplying them as props if they would be variable, but in this case they're quite static, no?

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
paulmskim: Same comment as above.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-header/template.js
paulmskim: I'll remove this, but I dislike using `defaultProps` for the reason that it's not as declarative and explicit as if you declared it like so. I know the `NumericLabel` component and how it works, so it's OK for me, but if a new dev came into this, it could easily get confusing.

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/attendees-action-button/template.js
mitogh: Right, I think this is because we need to make sure all strings are translatable afterwards, so we should address this one and the other one as well, will take care of that one in other PR, thanks.

plugins/tickets/src/modules/blocks/rsvp/rsvp-action-buttons/settings-action-button/template.js
mitogh: Perfect

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
mitogh: Makes sense, thanks for clarifying

plugins/tickets/src/modules/blocks/rsvp/template.js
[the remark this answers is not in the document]
mitogh: Right!

The main point here to have all strings as props in components and blocks to make sure strategy or anyone else can do a quick edit into the default props of a component or the object of the block without to much overhead.

I think this also removes hard code strings / variables and allow an external caller to change this if required at some point.

I think this one summarize the main point for the rest of the questions related with strings.

Let me know if you have additional questions, thank you.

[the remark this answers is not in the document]
paulmskim: I think it's safe to say that these strings will be set as they are and won't be changing, if there comes a time where it will change I or someone else can implement then. However, regarding your first comment, if I'm understanding you correctly, your reasoning to have these strings as props here is to make sure strategy can come in and change the code?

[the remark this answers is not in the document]
mitogh: That's right. And having all the strings as defaultProps of components makes easier to just find the right place of the string plus the component is a bit agnostic of what strings are used inside they just render the value of it, and we can get ride of the `__` dependency if the string is provided by the `container` instead so templates can remove references to those dependencies.

plugins/tickets/src/modules/blocks/rsvp/rsvp-container-content/template.js
mitogh: Yes, they are static, but relates to the point mentioned above.

plugins/tickets/src/modules/blocks/rsvp/template.js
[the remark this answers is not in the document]
mitogh: I think we can do that as long as those are provided by the connector. But overall the point is to make sure components just render stuff when they are stateless in this case using props to do so to quickly identify or change internals of any component is ideal.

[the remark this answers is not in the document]
mitogh: I think there a couple of additional things we can consider as advantages by just taking the string as a prop: 

1. Flexibility allow us to change this without change the need to change of the component, not common case but allow us to that at any point.
1. Consistency we define all strings as props of any component so is accesible from the outside world. 
1. Every plugin has a different domain name and if we ensure all strings are defined as props we can easily identify where strings are placed
1. Documentation, we can take advantage of scripts to document all the components with the available props making them clear on an easy to understand all the available props
1. Components are more pure as they are just render data and any external factor is isolated in props.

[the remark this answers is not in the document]
paulmskim: The basic definition of props from a container is to pass in data that is expected to change or functions that are expected to change the data. The static strings here are not props, they are not expected to change, and we are not listening for actions that will change these strings. A good example of where we can expect strings to change is in an element with custom separators, this is a great case for what you're talking about. But not at the block level.

Now, purer components is a great aspiration to aim for, but at what cost? You could say everything here can potentially change, and we should be as flexible as we can. What if we want our wrapper to be a `<section>` instead? What if we want our label to be a `<a>` tag? What if? Now we're looking at a "pure" component with 50 different props just because of "what if". There's a reasonable balance we have to achieve with props, flexibility vs specificity, and what we expect to happen.

Sorry, but I disagree here, the use case for static strings in block-level components as props is not the right approach. We can talk about it with the relevant people, and I would love to align on this, but I strongly believe this is the wrong direction to go.

You did, however, touch on consistency, and there is a case for having all strings defined in one place. In many large apps, passing strings as one large object to the global and accessing that object to get the required string is one common pattern. We can, and should, talk about how we plan to manage strings for this application, especially given that this project will continually grow.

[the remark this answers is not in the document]
mitogh: Yeah I think individually we might have different opinions on this one, but I think shouldn't be a blocker for now let's add this as an item to improve for the rest of the blocks as well as I said having a unified way to handle strings would be great, I think Loxi uses a JSON file, but I think we might require discussion around this and is not something critical at this point.

PR #318, opened 2018-09-16
Feature/112956 rsvp inactive state

paulmskim:
For ticket: https://central.tri.be/issues/112956

Depends on https://github.com/moderntribe/events-gutenberg/pull/315 and https://github.com/moderntribe/events-gutenberg/pull/317 and requires them to be merged first. I suggest merging the other PRs first before reviewing this as the diffs will show many that are the same.


mitogh:
Adding `hold` until dependencies are merged, for cleaner diff.

mitogh:

Thanks.

plugins/tickets/src/modules/blocks/rsvp/template.js
```diff
@@ -18,13 +18,28 @@ import {
...
-import { ActionDashboard } from '@moderntribe/tickets/elements';
+import { ActionDashboard, InactiveBlock } from '@moderntribe/tickets/elements';
+import { RSVPInactive } from '@moderntribe/tickets/icons';
 import './style.pcss';
 
 const RSVP = ( {
 	created,
 	isSelected,
 } ) => {
+	const inactiveBlockData = {
```
mitogh: Seems like this one does not require any additional prop I think this one can be placed outside of the block render method instead

PR #319, opened 2018-09-17
Add form to create a new ticket

mitogh:
Add UI for the form required to create a new ticket inside of the tickets blocks, separate components into smaller
parts to decrease complexity and have smaller tests.

Components part of the ticket form: 

- `<Header />`
- `<Capacity />`
- `<AdvancedOptions />`
- `<AttendeesRegistration />`


plugins/tickets/src/modules/elements/ticket-form/attendees-registration/template.js
```diff
@@ -0,0 +1,48 @@
...
+import './style.pcss';
+
+const AttendeesRegistration = ( props ) => {
+	const {
+		attendeeRegistrationLabel,
+		attendeeButtonLabel,
+		attendeeButtonClick,
+	} = props;
+
+	return (
```
paulmskim: this is OK for now, but in https://github.com/moderntribe/events-gutenberg/pull/322 I added a component called `labelWithModal`, which you can call like so:

https://github.com/moderntribe/events-gutenberg/pull/322/files#diff-954eb9025f65e7bcf8ec7eb67c9927b6R24

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,153 @@
...
+	selected: PropTypes.string,
+	onSelect: PropTypes.func.isRequired,
+	options: PropTypes.arrayOf( PropTypes.shape( {
+		name: PropTypes.string,
+		value: PropTypes.string,
+	} ) ).isRequired,
+};
+
+// Custom input for this type of form
+const Input = ( { type, input, selected } ) => (
```
paulmskim: doesn't have to be in this PR, but would be nice to make this a general component as there are many pieces that can be reused.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,153 @@
...
+		type,
+		onSelectType,
+		independent,
+		shared
+	} = props;
+	return (
+		<div className="tribe-editor__tickets-form__row">
+			<div className="tribe-editor__tickets-form__labels">
+				<LabelWithTooltip
+					id="tickets-capacity-type"
```
paulmskim: This will not work well once you have 2 `ticket` blocks, which we will definitely see. We cannot have 2 or more elements with the same id. Having the same id for all ticket capacity selects will cause problems on the page.

Instead, I suggest 2 things:
- Make this a stateful component by extending `PureComponent`, create a property `this.selectId = uniqid()` in the `constructor`, and add it to the `Select` below.
- As mentioned above with the `LabelWithTooltip` element, as this label and tooltip reference a group of form elements, it's probably better to have a label associated with the select itself, but visually hidden. So something like:

```js
...
  <label htmlFor={ this.selectId } >...</label>
  <Select
    id={ this.selectId }
    ...
  />
...
```

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
```diff
@@ -0,0 +1,15 @@
+.tribe-editor__ticket-duration__duration-picker {
+	position: relative;
+	margin-bottom: 20px;
+
+	.tribe-editor__date-time-range-picker__start {
+		width: 100%;
+		margin-left: 0;
+	}
+
+	.tribe-editor__date-time-range-picker__separator--time-range {
```
paulmskim: wonder why this style is needed? doesn't the date time range picker style not have the styles set already?

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
```diff
@@ -0,0 +1,15 @@
+.tribe-editor__ticket-duration__duration-picker {
+	position: relative;
+	margin-bottom: 20px;
+
+	.tribe-editor__date-time-range-picker__start {
```
paulmskim: this style should already be taken care of in the date time range picker element. if this is to align the date time range picker with the label + tooltip, check out these:

- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration/template.js#L14
- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration/style.pcss#L1
- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration-label/style.pcss#L1
- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration-picker/template.js
- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration-picker/style.pcss#L1

If you have question let me know and we can check it out together, the elements should really just be plug-and-play.

plugins/tickets/src/modules/elements/ticket-form/duration/template.js
```diff
@@ -0,0 +1,56 @@
...
+ * Internal dependencies
+ */
+import { DateTimeRangePicker, LabelWithTooltip } from '@moderntribe/tickets/elements';
+import './style.pcss';
+
+const TicketDurationPicker = ( props ) => {
+	const { label, tooltip } = props;
+	return (
+		<div className="tribe-editor__tickets-form__row">
+			<div className="tribe-editor__tickets-form__labels">
```
paulmskim: as in the previous comment, we shouldn't need wrapper divs for the label and date time range picker, these elements should just be plug and play, check out:

- https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/tickets/src/modules/blocks/rsvp/rsvp-duration/template.js#L14

plugins/tickets/src/modules/elements/ticket-form/advanced-options/template.js
```diff
@@ -0,0 +1,82 @@
...
+
+	static defaultProps = {
+		accordionId: 'ticketsPlaceholder',
+		contentId: 'ticketsPlaceholder',
+		headerId: 'ticketsPlaceholder',
+		isActive: false,
+		accordionTitle: __( 'Advanced Options', 'events-gutenberg' ),
+		onClick: () => {},
+	};
+
```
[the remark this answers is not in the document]
paulmskim: **EDIT**: I didn't realize this was actually an element. This really isn't an element, it's specific for the `ticket` block. I cannot reuse this for RSVP, and there won't be any other blocks/components that would use this element, it's too specific. All of this code under the `ticket-form` should actually be under the `ticket` block as its own components.

plugins/tickets/src/modules/elements/ticket-form/footer/style.pcss
```diff
@@ -0,0 +1,28 @@
...
+	&:hover {
+		color: #000;
+	}
+}
+
+.tribe-editor__btn--label.tribe-editor__tickets-form__btn-create {
+	background-color: #009FD4;
+	color: #FFF;
+
+	&:hover {
```
paulmskim: should have `&:focus` here as well. however, i have a PR currently open that addresses us styling this button over and over again:

https://github.com/moderntribe/events-gutenberg/pull/320/files#diff-45709bc27430fe9a9fb02d309e4e1264R18

we should be using the class `tribe-editor__button--sm` for everything where we have the small blue button from the style guide.

plugins/tickets/src/modules/elements/ticket-form/footer/style.pcss
```diff
@@ -0,0 +1,28 @@
...
+		background-color: #027BB4;
+		color: #FFF;
+	}
+
+	&:disabled {
```
paulmskim: you can simplify this by doing:

```
&:disabled {
  &,
  &:hover,
  &:focus {
    /* your styles go here */
  }
}
```

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,115 @@
...
+	const priceLabel = [ <span key="price-currency">{priceCurrency}</span>, priceInput ];
+
+	return (
+		<header className="tribe-editor__new-ticket__header">
+			<div className="tribe-editor__new-ticket__icon">
+				<TicketStatus expires={ expires } />
+			</div>
+			<div className="tribe-editor__new-ticket__content">
+				<AutosizeInput
+					id="ticket-creation-title"
```
paulmskim: this should be using a unique id. it's likely there won't be 2 ticket creations happening at the same time, but we cannot have 2 of the same id on the page.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,115 @@
...
+					name="ticket-creation-title"
+					className="tribe-editor__new-ticket__title"
+					value={ title }
+					placeholder={ titlePlaceholder }
+					onChange={ sendValue( setTitle ) }
+				/>
+				<AutosizeInput
+					id="ticket-creation-description"
```
paulmskim: same comment as above on the id

plugins/tickets/src/modules/elements/ticket-form/header/style.pcss
```diff
@@ -0,0 +1,99 @@
+.tribe-editor__new-ticket__header {
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	border: 1px solid #AEB4BB;
+	border-bottom: none;
+
+	&:after {
```
paulmskim: we should not be using spacer elements, even if they are pseudo elements, instead, use a margin to the right of `.tribe-editor__new-ticket__price`.

plugins/tickets/src/modules/elements/ticket-form/header/style.pcss
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+.tribe-editor__new-ticket__content {
+	flex: 1;
+	padding-left: 28px;
+	display: flex;
+	flex-wrap: wrap;
+
+	.tribe-editor__new-ticket__title,
```
paulmskim: these shouldn't need to be under `.tribe-editor__new-ticket__content` as we are unnecessarily increasing the specificity of these elements, the BEM should put it in its own 'namespace', so to speak, and should be enough.

plugins/tickets/src/modules/elements/ticket-form/sku/template.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { sendValue } from '@moderntribe/common/utils/input';
+import { LabelWithTooltip } from '@moderntribe/tickets/elements';
+
+const SKU = ( { onChange, value, label, tooltip } ) => (
+	<div className="tribe-editor__tickets-form__row">
+		<div className="tribe-editor__tickets-form__labels">
```
paulmskim: not wrong, but these shouldn't need these wrapper divs.

plugins/tickets/src/modules/elements/ticket-form/sku/template.js
```diff
@@ -0,0 +1,53 @@
...
+			<LabelWithTooltip
+				id="ticket-sku"
```
paulmskim: these ids should be unique, as mentioned in previous comments.

plugins/tickets/src/modules/elements/ticket-form/sku/template.js
```diff
@@ -0,0 +1,53 @@
...
+				label={ label }
+				tooltipText={ tooltip }
+				tooltipLabel={ <Dashicon icon="info-outline" /> }
+			/>
+		</div>
+		<div className="tribe-editor__tickets-form__input-group">
+			<input id="ticket-sku" type="text" value={ value } onChange={ sendValue( onChange ) } />
```
paulmskim: unique id

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,153 @@
...
+					tooltipText={ capacityToolTip }
+					tooltipLabel={ <Dashicon icon="info-outline" /> }
+				/>
+			</div>
+			<div className="tribe-editor__tickets-form__input-group">
+				<Select
+					selected={ type }
+					options={ capacityOptions }
+					onSelect={ onSelectType }
+					id="tickets-capacity-type"
```
paulmskim: same thing with unique id

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+// Custom input for this type of form
+const Input = ( { type, input, selected } ) => (
+	type === selected && (
+		<div className="tribe-editor__tickets-form__input-row">
+			<label htmlFor={ `ticket-input-${ type }` }>{ input.label }</label>
+			<input
+				type="number"
+				id={ `ticket-input-${ type }` }
```
paulmskim: This id needs to be unique. In the `Input` elements below, they are just passing in a `type`, which creates 2 inputs with the same id. This is also the case with multiple ticket blocks, and you have multiple elements with the same id.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,115 @@
...
+		priceCurrency,
+		pricePlaceholder,
+		setPrice,
+		expires,
+	} = props;
+
+	const priceInput = (
+		<AutosizeInput
+			key="price-input"
+			id="ticket-creation-description"
```
paulmskim: unique id

plugins/tickets/src/modules/blocks/tickets/settings/template.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+import {
+	CapacityTable,
+	HeaderImage,
+} from '@moderntribe/tickets/elements';
+
+const Settings = () => (
+	<Fragment>
+		<CapacityTable />
+		<HeaderImage />
```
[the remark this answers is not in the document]
mitogh: For now is just a placeholder with no toggle capability, mainly to group all this elements once that one is available.

plugins/tickets/src/modules/blocks/tickets/style.pcss
```diff
@@ -2,3 +2,34 @@
 	color: #545D66;
 	font-family: Helvetica, sans-serif, arial;
 }
+
```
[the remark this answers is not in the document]
mitogh: Seems like was added after the merge conflicts will double check.

plugins/tickets/src/modules/elements/accordion-header/element.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Wordpress dependencies
+ */
+import { Dashicon } from '@wordpress/components';
+
+const AccordionHeader = ( { active, title, iconClass, titleClass } ) => (
```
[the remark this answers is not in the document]
mitogh: Right however the idea is to remove the one from RSVP instead and use this one instead to decrease the amount of components with the same behavior across multiple elements.

plugins/tickets/src/modules/elements/label-with-tooltip/element.js
```diff
@@ -11,20 +11,38 @@ import classNames from 'classnames';
 import { Tooltip } from '@moderntribe/common/elements';
 import './style.pcss';
 
+const Label = ( { id, label } ) => {
```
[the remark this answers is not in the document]
mitogh: I think this one of the things we discuss on the call yesterday to split this one in order to render a `<label>` instead.

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
mitogh: Styles are applied to the container instead as we have an element that is absolute placed inside of this container as a reference.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
mitogh: Great, not sure where the scenario of 2 ticket blocks comes into place tho would you mind explain that one a bit more?

plugins/tickets/src/modules/elements/ticket-form/attendees-registration/template.js
mitogh: Well I think we are creating components to specifc at this point at least based on names instead we should compose them as: 

- Label -> tooltip( label ) -> modal( tooltip( label ) )

As what we are doing at this point is: 

- Label
- LabelWithToolTip
- LabelWithModal 

First approach seems more scalable, however I haven't look into the labelWithTooltip but seems like we should do a composition instead of specificity on those cases.

plugins/tickets/src/modules/elements/accordion-header/element.js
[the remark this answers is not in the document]
paulmskim: As we build elements on top of each other to create components and blocks, things become more and more specific. At the block level is where specificity is supposed to be, not at the element level. The elements are supposed to provide a generic template and interface for which to build blocks with, which provide their own specific requirements.

Like I said, technically speaking this isn't wrong, but semantically speaking, this is way too specific for an element.

plugins/tickets/src/modules/elements/label-with-tooltip/element.js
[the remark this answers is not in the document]
paulmskim: So I don't think I understood you when you said you were going to use a label. I'm pretty sure when we discussed this on the call, I was of the mind that using a label within this component is not the best idea.

As mentioned above, I don't think using a label component is best here. I'd love to discuss if you think otherwise.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
paulmskim: within the `tickets` block you can have many `ticket` blocks.

plugins/tickets/src/modules/elements/ticket-form/attendees-registration/template.js
paulmskim: I think you might be misunderstanding what the modal portion is, it's actually not a wrapper around the tooltip, it's a replacement of the tooltip. So if you look into that PR, we have a wrapper element called `LabeledItem` which wraps the `children` passed into the `LabeledItem` component with a label next to it. This seems to be a recurring pattern with the latest being the `Tooltip` and `ModalButton`. The labeled elements `LabelWithTooltip` and `LabelWithModal` created from the `Tooltip` and `ModalButton` elements are use cases of `LabeledItem`, which still provide a lot of flexibility (label text, modal button text, modal button functionality, tooltip text, tooltip label, etc.), so I disagree that this is too specific. In fact, these elements (`LabelWithTooltip`, `LabelWithModal`) are composed elements themselves, they just have a name to them.

plugins/tickets/src/modules/elements/ticket-form/footer/style.pcss
mitogh: Curios how we set the `disabled` and `focus` style here if both are different.

plugins/tickets/src/modules/elements/ticket-form/footer/style.pcss
mitogh: Great adding a TODO here instead to remove this and update for a global style instead.

plugins/tickets/src/modules/elements/ticket-form/attendees-registration/template.js
mitogh: Haven't look into the specifics on this one but is my interpretation as we have similar names as mentioned above.

plugins/tickets/src/modules/elements/label-with-tooltip/element.js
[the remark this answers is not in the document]
mitogh: I think this will prevent us to create another component just to use a `<label>` tag instead, there might be cases when just a `<span>` is good but as long as you pass and ID to associated with the `<label>` the component will use a `<label>`. 

Probably create another component for this seems to more complex from my point of view as both are going to share same structure only difference is one is using `<span>` and other `<label>`

plugins/tickets/src/modules/elements/accordion-header/element.js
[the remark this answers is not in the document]
mitogh: In this particular case I think it does makes sense as you are creating the same `header` for both different containers, unless we have a single container for `accordions` as both `accordions` share the same structure in design and markup. 

I think having a single one will help to have similar patterns from design group as components just like the `small btn` as is the same one used in multiple places, the same style and structure for the header is used in this two accordions.

plugins/tickets/src/modules/elements/ticket-form/advanced-options/style.pcss
```diff
@@ -0,0 +1,44 @@
+.tribe-editor__tickets__advanced-options {
+	.tribe-editor__accordion__row-header {
+		background-color: transparent;
```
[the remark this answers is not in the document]
mitogh: I think this one relates with the one above when we have the styles in multiple places for the same style of component so I think having a single style for the component would be ideal.

Will update the styles of this one in the location of the component, instead.

plugins/tickets/src/modules/elements/accordion-header/element.js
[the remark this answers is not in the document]
mitogh: Looking into more detail I think you're right, removing this one 👍

plugins/tickets/src/modules/elements/ticket-form/advanced-options/template.js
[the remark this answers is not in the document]
mitogh: I realize this after I check on your side, but wanted to place all elements to avoid having nested or specifc ones in directories of each block.

I'm fine with moving them into the `tickets` directory, I think Neal was using `components/` instead of `elements` I think we should define how this should be approached, but I do agree with the part of `elements` should contain only reusable items.

Will create a different PR to avoid more complexity on this one after merging this one by moving all this inside of the `tickets` block.

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
mitogh: I think the main difference here is the structure of the layout of the inputs as those in the tickets have the same width but the `to` is placed outside of that container.

So we need to make sure the inputs on those fill all the available space on that row, and place the `to` outside of that container.

- http://p.tri.be/7mfKOO

Let me know if you have more questions in particular about that one.

mitogh:
I think I've addressed all the comments and answer the rest, thanks for such a great review.

Let me know if you have more concerns / questions to be addressed before moving this one into the `feature/` branch.

Thanks @paulmskim

plugins/tickets/src/modules/elements/ticket-form/advanced-options/template.js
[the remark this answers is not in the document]
paulmskim: OK, the sooner the better, as we should architecturally have things in place first.

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
paulmskim: If you just plug in the `<DateTimeRangePicker />` component with no external styling, you will see that that styling is already take care of. Check out below if you want more details on the markup and styles:

https://github.com/moderntribe/events-gutenberg/tree/feature/event-ticket-blocks/plugins/tickets/src/modules/elements/date-time-range-picker

These elements really are plug-and-play, not a ton of styles changes needed, only layout.

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
paulmskim: This is not resolved, this relates to https://github.com/moderntribe/events-gutenberg/pull/319#discussion_r218719427

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -42,7 +43,7 @@ const Header = ( props ) => {
 	const priceInput = (
 		<AutosizeInput
 			key="price-input"
-			id="ticket-creation-description"
+			id={ uniqid( 'ticket-creation-price-' ) }
```
paulmskim: this component needs to be a stateful component that extends `PureComponent`, much like you did here:

https://github.com/moderntribe/events-gutenberg/pull/319/commits/ea10a086c07e9d1f1a0a33ad6e7ff5c7624e49fd

If any of these props change, because this component is stateless, the entire component will rerender, which in turn will call `uniqid()` again for every instance, recreating a new unique id.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -59,15 +60,15 @@ const Header = ( props ) => {
 			</div>
 			<div className="tribe-editor__new-ticket__content">
 				<AutosizeInput
-					id="ticket-creation-title"
+					id={ uniqid( 'ticket-creation-title-' ) }
```
paulmskim: same on the id comment above

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -59,15 +60,15 @@ const Header = ( props ) => {
...
 					name="ticket-creation-title"
 					className="tribe-editor__new-ticket__title"
 					value={ title }
 					placeholder={ titlePlaceholder }
 					onChange={ sendValue( setTitle ) }
 				/>
 				<AutosizeInput
-					id="ticket-creation-description"
+					id={ uniqid( 'ticket-creation-description-' ) }
```
paulmskim: same on the id comment above

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -42,19 +43,22 @@ Select.propTypes = {
...
-				type="number"
-				id={ `ticket-input-${ type }` }
-				value={ input.value }
-				onChange={ input.onChange }
-			/>
-		</div>
-	)
-);
+const Input = ( { type, input, selected } ) => {
+	const id = uniqid( `ticket-input-${ type }-` );
```
paulmskim: we should be passing in the id as a prop instead of defining it within the component. also, if any props change, this entire component will rerender, creating a new unique id.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -77,11 +81,13 @@ const Capacity = ( props ) => {
 		independent,
 		shared
 	} = props;
+
+	const selectID =  uniqid( 'capacity-type-' );
```
paulmskim: this `Capacity` component should be a stateful component extending `PureComponent`, much like you did here: https://github.com/moderntribe/events-gutenberg/pull/319/commits/ea10a086c07e9d1f1a0a33ad6e7ff5c7624e49fd

plugins/tickets/src/modules/elements/ticket-form/sku/template.js
```diff
@@ -10,28 +10,32 @@ import noop from 'lodash/noop';
...
-	<div className="tribe-editor__tickets-form__row">
-		<div className="tribe-editor__tickets-form__labels">
-			<LabelWithTooltip
-				id="ticket-sku"
-				label={ label }
-				tooltipText={ tooltip }
-				tooltipLabel={ <Dashicon icon="info-outline" /> }
-			/>
+const SKU = ( { onChange, value, label, tooltip } ) => {
+	const id = uniqid( 'ticket-sku' );
```
paulmskim: same note on the unique id as above. this needs to extend `PureComponent`.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,116 @@
...
+		priceCurrency,
+		pricePlaceholder,
+		setPrice,
+		expires,
+	} = props;
+
+	const priceInput = (
+		<AutosizeInput
+			key="price-input"
+			id={ uniqid( 'ticket-creation-price-' ) }
```
paulmskim: the `Header` component needs to be a stateful component that extends `PureComponent`, much like you did here: https://github.com/moderntribe/events-gutenberg/pull/319/commits/ea10a086c07e9d1f1a0a33ad6e7ff5c7624e49fd

If any of these props change, because this component is stateless, the entire component will rerender, which in turn will call `uniqid()` again for every instance, recreating a new unique id.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,116 @@
...
+	const priceLabel = [ <span key="price-currency">{priceCurrency}</span>, priceInput ];
+
+	return (
+		<header className="tribe-editor__new-ticket__header">
+			<div className="tribe-editor__new-ticket__icon">
+				<TicketStatus expires={ expires } />
+			</div>
+			<div className="tribe-editor__new-ticket__content">
+				<AutosizeInput
+					id={ uniqid( 'ticket-creation-title-' ) }
```
paulmskim: same comment as above on `uniqid` and stateful component.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
```diff
@@ -0,0 +1,116 @@
...
+					name="ticket-creation-title"
+					className="tribe-editor__new-ticket__title"
+					value={ title }
+					placeholder={ titlePlaceholder }
+					onChange={ sendValue( setTitle ) }
+				/>
+				<AutosizeInput
+					id={ uniqid( 'ticket-creation-description-' ) }
```
paulmskim: same as above on `uniqid` and stateful component.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,158 @@
...
+	onSelect: PropTypes.func.isRequired,
+	options: PropTypes.arrayOf( PropTypes.shape( {
+		name: PropTypes.string,
+		value: PropTypes.string,
+	} ) ).isRequired,
+};
+
+// Custom input for this type of form
+const Input = ( { type, input, selected } ) => {
+	const id = uniqid( `ticket-input-${ type }-` );
```
paulmskim: we should be passing in the id as a prop instead of defining it within the component. also, if any props change, this entire component will rerender, creating a new unique id.

plugins/tickets/src/modules/elements/ticket-form/capacity/template.js
```diff
@@ -0,0 +1,158 @@
...
+		capacityLabel,
+		capacityToolTip,
+		capacityOptions,
+		type,
+		onSelectType,
+		independent,
+		shared
+	} = props;
+
+	const selectID =  uniqid( 'capacity-type-' );
```
paulmskim: this `Capacity` component should be a stateful component extending `PureComponent`, much like you did here: https://github.com/moderntribe/events-gutenberg/pull/319/commits/ea10a086c07e9d1f1a0a33ad6e7ff5c7624e49fd

plugins/tickets/src/modules/elements/ticket-form/sku/template.js
```diff
@@ -0,0 +1,56 @@
...
+import uniqid from 'uniqid';
+
+/**
+ * Internal dependencies
+ */
+import { sendValue } from '@moderntribe/common/utils/input';
+import { LabelWithTooltip } from '@moderntribe/tickets/elements';
+
+const SKU = ( { onChange, value, label, tooltip } ) => {
+	const id = uniqid( 'ticket-sku' );
```
paulmskim: same note on the unique id as above. this needs to extend `PureComponent`.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
mitogh: Right, however in this case the new ID does not make any difference as is just an associated with the label, so having a different ID will not have any difference here as having the same as before.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
mitogh: Same applied here.

plugins/tickets/src/modules/elements/ticket-form/header/template.js
mitogh: Would you mind take a look at this @paulmskim ?

paulmskim:
@mitogh still have a couple here: https://github.com/moderntribe/events-gutenberg/pull/319#discussion_r218719427 and https://github.com/moderntribe/events-gutenberg/pull/319#discussion_r218719552

also left you a note on https://github.com/moderntribe/events-gutenberg/pull/319#discussion_r218932676

mitogh:
I think all items has been addressed.

Thanks and let me know if you have more questions.

plugins/tickets/src/modules/elements/ticket-form/duration/style.pcss
paulmskim: The RSVP block has exactly the same feature, which is why I'm saying these styles are unnecessary. As mentioned before, these elements should just be used as is, and fix the outer element layouts as necessary without touching the inner element layout (which should be handled in the element styles). However, ass it is, it's not functionally a blocker, so I will leave it to you to decide if you would like to fix it or not.

PR #322, opened 2018-09-18
Feature/112925 attendee registration

paulmskim:
For ticket: https://central.tri.be/issues/112925

Depends on https://github.com/moderntribe/events-gutenberg/pull/317 and requires it to be merged first. I suggest merging the other PR first before reviewing this as the diffs will show many that are the same.


paulmskim:
I don't know why tests are failing here, they're passing on my local.

mitogh:

Thanks.

plugins/common/src/modules/elements/accordion/AccordionRow/template.js
```diff
@@ -11,90 +11,112 @@ import classNames from 'classnames';
...
+			'aria-controls': this.contentId,
+			'aria-expanded': _isActive,
+			'aria-selected': _isActive,
+			id: this.headerId,
+			role: 'tab',
+		};
+	};
+
+	getContentAttrs = () => {
+		return {
```
mitogh: `return` is not required here

plugins/common/src/modules/elements/accordion/AccordionRow/template.js
```diff
@@ -11,90 +11,112 @@ import classNames from 'classnames';
...
-	isActive: PropTypes.bool.isRequired,
-	onClick: PropTypes.func,
-	onClose: PropTypes.func,
-	onOpen: PropTypes.func,
-};
+		return (
+			<article
+				className={ classNames(
+					'tribe-editor__accordion__row',
+					{ 'active': this.state.isActive },
```
mitogh: Quotes are `optional` here for `active`

plugins/common/src/modules/elements/label-with-modal/__tests__/element.test.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import LabelWithModal from '@moderntribe/common/elements/label-with-modal/element';
+
+jest.mock( '@wordpress/components', () => ( {
+	Modal: ( { title, children } ) => (
```
mitogh: I think we should add into the global mocks in case `Modal` is ever tested somewhere or when a component using `LabelWithModal`  is tested.

plugins/tickets/src/modules/blocks/rsvp/rsvp-attendee-registration/template.js
```diff
@@ -0,0 +1,34 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { LabelWithModal } from '@moderntribe/common/elements';
+import './style.pcss';
+
+const RSVPAttendeeRegistration = () => {
+	const closeButtonLabel = __( 'close', 'events-gutenberg' );
```
mitogh: Seems like this 3 either can be props or placed outside of this component that will simplify the render method of this one

plugins/common/src/modules/elements/labeled-item/element.js
```diff
@@ -0,0 +1,30 @@
...
+const LabeledItem = ( {
+	className,
+	label,
+	children,
+} ) => (
+	<div className={ classNames(
+		'tribe-editor__labeled-item',
+		className,
+	) }>
+		<span className="tribe-editor__labeled-item__label">
```
mitogh: I think `<span>` should be rendered only if `label` is present

plugins/common/src/modules/elements/modal-button/element.js
```diff
@@ -0,0 +1,88 @@
...
+				'tribe-editor__modal-button',
+				className,
+			) }>
+				<Button
+					className="tribe-editor__modal-button__button"
+					onClick={ this.onClick }
+				>
+					{ label }
+				</Button>
+				{ this.state.isOpen && (
```
mitogh: Seems like a good one to split into his own method and call here `{ this.renderModal() }`

plugins/common/src/modules/elements/modal-button/element.js
```diff
@@ -0,0 +1,88 @@
...
+		this.props.onCloseClick && this.props.onCloseClick( e );
+		this.onRequestClose();
+	}
+
+	onRequestClose = () => {
+		this.onClose();
+		this.setState( { isOpen: false } );
+	}
+
+	onOpen = () => this.props.onOpen && this.props.onOpen();
```
mitogh: I think to avoid this check we can take advantage of `isRequired` instead

plugins/common/src/modules/elements/modal-button/element.js
```diff
@@ -0,0 +1,88 @@
...
+
+	onClose = () => this.props.onClose && this.props.onClose();
```
mitogh: Same as above.

plugins/common/src/modules/elements/modal-button/element.js
```diff
@@ -0,0 +1,88 @@
...
+
+	constructor( props ) {
+		super( props );
+		this.state = {
+			isOpen: false,
+		};
+	}
+
+	onClick = ( e ) => {
+		this.props.onClick && this.props.onClick( e );
```
mitogh: Making this one required will prevent this check

plugins/common/src/modules/elements/modal-button/__tests__/element.test.js
```diff
@@ -0,0 +1,62 @@
...
+import React from 'react';
+import { noop } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import ModalButton from '@moderntribe/common/elements/modal-button/element';
+
+jest.mock( '@wordpress/components', () => ( {
+	Modal: ( { title, children } ) => (
```
mitogh: Having this one into the globals mocks will prevent to do this twice

plugins/common/src/modules/elements/accordion/AccordionRow/__tests__/template.test.js
```diff
@@ -21,85 +16,40 @@ describe( 'Accordion Row Element', () => {
...
-			const attributes = getContentAttrs( row );
-			expect( attributes ).toMatchSnapshot();
-		} );
+	it( 'executes onOpen and onClose handlers', ( done ) => {
+		const component = mount( <AccordionRow { ...row } /> );
+		component.find( 'button' ).simulate( 'click' );
+		setTimeout( () => {
+			component.find( 'button' ).simulate( 'click' );
+		}, 250 );
+		setTimeout( () => {
```
mitogh: Wonder if we should use `jest.setTimeout` instead of global one or ` jest.useFakeTimers()`

plugins/common/src/modules/elements/labeled-item/element.js
paulmskim: if you're using the `LabeledItem` element, then you definitely should have a `label`. Otherwise, there's no point of using the labeled item, you might as well just have the item.

plugins/common/src/modules/elements/modal-button/element.js
paulmskim: There are cases where we do not have an `onOpen` function passed in, so requiring a function to be passed in just to avoid a check is more work on the upper levels. As we talked about earlier, elements should be flexible, and this is allowing for flexibility of both passing in a function or not. Passing in no function here is passing in `undefined`, which means it won't run.

plugins/common/src/modules/elements/accordion/AccordionRow/__tests__/template.test.js
paulmskim: `jest.setTimeout` sets the default timeout value for each test, it doesn't do what you think it does. `jest.useFakeTimers` mocks a `setTimeout` or some other timed event that actually exists in the code.

plugins/common/src/modules/elements/modal-button/element.js
mitogh: And how about to define a default prop as `noop` instead to avoid the need to set the value as required that also will remove the need to add an extra conditional.

mitogh:
Seems like tests are not passing either I think you might need to rerun travis to make it pass

plugins/common/src/modules/elements/modal-button/element.js
paulmskim: Having checks aren't bad. Semantically, doing `isRequired` and `noop` doesn't make sense but programmatically it works. I'll make the change.

mitogh: Great, not necessarily because it's bad but I think code can be simpler by using any of those mechanism.

bordoni: I feel like we don't need to worry about those changes now @mitogh

PR #323, opened 2018-09-19
Allow for saving on new event creation

mitogh:
Make sure when the event is first created the diff is not generated
until the save has been saved.


mitogh:
Would you mind take another look: 

@bordoni @juanfra @paulmskim 

As there was a conflict

PR #325, opened 2018-09-19
Fix/114451 container panel element fix

paulmskim:
For ticket: https://central.tri.be/issues/114451

Depends on https://github.com/moderntribe/events-gutenberg/pull/317 getting merged first.


mitogh:

Thanks.

plugins/tickets/src/modules/elements/container-panel/element.js
```diff
@@ -10,37 +10,59 @@ import classNames from 'classnames';
  */
 import './style.pcss';
 
+export const RSVP = 'rsvp';
```
mitogh: I think we should export this one as a single constant like: 

```js
export const LAYOUTS = {
	ticket: 'ticket',
	rsvp: 'rsvp', 
}
```

plugins/tickets/src/modules/elements/container-panel/element.js
```diff
@@ -10,37 +10,59 @@ import classNames from 'classnames';
...
+				{ header }
+			</div>
+			<div className="tribe-editor__container-panel__content">
+				{ content }
+			</div>
+		</Fragment>
+	);
+
+	const getHeaderAndContent = () => (
+		layout === TICKET
```
mitogh: So we can make this clear here and external places into `LAYOUTS.ticket`

plugins/tickets/src/modules/elements/container-panel/element.js
```diff
@@ -10,37 +10,59 @@ import classNames from 'classnames';
...
 
 ContainerPanel.propTypes = {
 	className: PropTypes.string,
-	leftSection: PropTypes.node,
-	rightSectionContent: PropTypes.node,
-	rightSectionHeader: PropTypes.node,
+	content: PropTypes.node,
+	header: PropTypes.node,
+	icon: PropTypes.node,
+	layout: PropTypes.oneOf( [ RSVP, TICKET ] ).isRequired,
```
mitogh: So this one can be converted into `Object.keys( LAYOUTS )`

plugins/tickets/src/modules/elements/container-panel/element.js
mitogh: This will require only this place to be updated if ever required.

PR #320, opened 2018-09-17
Feature/112931 settings panel

paulmskim:
For ticket: https://central.tri.be/issues/112931

Depends on https://github.com/moderntribe/events-gutenberg/pull/317 and requires it to be merged first. I suggest merging the other PR first before reviewing this as the diffs will show many that are the same.

A couple things here:
- `Button` component styles have been added as default to the button. The style guide button style for small button has also been added, use the class `tribe-editor__button--sm` to get the style.
- Renamed `HeaderImage` to `ImageUpload` and moved to `common` plugin to make it more generic.
- See `RSVPDashboard` component to see how to hook up the `ActionDashboard` and `SettingsDashboard` together. This will has bearings on the `Tickets` block.

PR #326, opened 2018-09-19
Feature/112936 header image upload

paulmskim:
For ticket: https://central.tri.be/issues/112936

Depends on https://github.com/moderntribe/events-gutenberg/pull/320


plugins/common/src/modules/elements/image-upload/element.js
```diff
@@ -26,32 +28,73 @@ export const renderMediaUpload = ( label ) => ( { open } ) => (
...
+	buttonLabel,
+	className,
+	description,
+	imageAlt,
+	imageSrc,
+	onRemove,
+	onSelect,
+	title,
+} ) => (
+	<div className={ classNames(
```
mitogh: I think at this point it might be better to split all the conditional renders into variables here to have the final return more clear.

plugins/common/src/modules/elements/image-upload/element.js
```diff
@@ -26,32 +28,73 @@ export const renderMediaUpload = ( label ) => ( { open } ) => (
 	</Button>
 );
 
-const ImageUpload = ( { title, description, buttonLabel, onSelect, mediaId } ) => (
-	<div className="tribe-editor__image-upload">
+export const renderImage = ( imageAlt, imageSrc, onRemove ) => (
```
mitogh: I think this one can be a component even is is declared on this component as will help to make code more clear in terms of we can replace: 

```js
renderImage( imageAlt, imageSrc, onRemove )
```

Into 


```js
<Image alt={imageAlt} src={imageSrc} onRemove={ onRemove } />
```

And params names can get ride of `image`  plus the level of effort to make a component seems fair enough

plugins/common/src/modules/elements/image-upload/element.js
```diff
@@ -26,32 +28,73 @@ export const renderMediaUpload = ( label ) => ( { open } ) => (
...
-				onSelect={ onSelect }
-				type="image"
-				value={ mediaId }
-				render={ renderMediaUpload( buttonLabel ) }
-			/>
+			{
+				imageSrc
+					? renderImage( imageAlt, imageSrc, onRemove )
+					: (
+						<MediaUpload
```
mitogh: I think we should provide the ID of the current selection if any for this one as well.

plugins/common/src/modules/elements/image-upload/element.js
```diff
@@ -26,32 +28,73 @@ export const renderMediaUpload = ( label ) => ( { open } ) => (
...
 		</div>
 	</div>
 );
 
 ImageUpload.propTypes = {
 	buttonLabel: PropTypes.string,
+	className: PropTypes.string,
 	description: PropTypes.string,
-	mediaId: PropTypes.number,
+	imageAlt: PropTypes.string,
```
mitogh: I think defining a single object as image shape here into: 

```js
image: {
	alt:
	src
}
```

Also provides flexibility to add extra props later on if required like `orientation` or image size like `width` or `height`

mitogh: And I think `alt` should be a required prop on that object

plugins/tickets/src/modules/blocks/rsvp/rsvp-header-image/template.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+import { ImageUpload } from '@moderntribe/common/elements';
+
+const RSVPHeaderImage = ( {
+	headerImageAlt,
+	headerImageSrc,
+	onRemove,
+	onSelect
+} ) => {
+	const imageUploadProps = {
```
mitogh: I think this one can be placed outside of this component

plugins/tickets/src/modules/blocks/rsvp/rsvp-header-image/container.js
```diff
@@ -0,0 +1,34 @@
...
+import { DEFAULT_STATE } from '@moderntribe/tickets/data/blocks/rsvp/reducer';
+import { withStore } from '@moderntribe/common/hoc';
+
+const mapStateToProps = ( state ) => ( {
+	headerImageAlt: selectors.getRSVPHeaderImageAlt( state ),
+	headerImageSrc: selectors.getRSVPHeaderImageSrc( state ),
+} );
+
+const mapDispatchToProps = ( dispatch ) => ( {
+	onRemove: () => dispatch( actions.setRSVPHeaderImage( {
```
mitogh: I think on this one having just a `header` property is be set on store makes things easier in multiple ways as: 

1. Removes the need to set multiple values if required
1. Set and remove a value is simpler as you just pass the value from the callback in this case `image` and to remove we can just set `null` into the store.

And in order to get access into the values we can use the selectors to the job to select the correct values from the `image` fields instead I think by having the complete `image` object in the store we can use any of their properties without the need to add / remove a new key into the reducer, and I think reducer keeps it simple as it remains as: 

`header: null // no value`
`header: { id: ... } // values present`

And to access values of the store of this object as mentioned above we can use `selectors` to do the heavy lifting and will allow to check for things when image.

By doing this we decrease the amount of code needed to handle this and the need of having to use a saga as it will require just to set a single value into the store.

plugins/common/src/modules/elements/image-upload/element.js
paulmskim: The way this element works, once an image is selected, the image renders and the media upload button disappears. There is no way for the user to get back to the media upload dialog box, so at this point it's not needed. The other reason why I don't want to pass it in here is because currently the state does not hold the `id`, only the `src` and `alt`. If I need to pass in the `id` here, that's another property I need to create for the state, actions, reducer cases, and selectors for something that doesn't even do anything.

plugins/common/src/modules/elements/image-upload/element.js
paulmskim: This function here would not become an `Image` component, it holds an `` as well as a `Button`, and that would semantically not make sense to call it an `Image`. However, I do see the case for `` becoming an `Image` component, but I purposely stayed away from creating it because of the huge amount of props that could be passed into the component. I will have another look at this.

bordoni:
On this one let's go with the ID storage to make it possible to this value available to the Frontend View of the blocks.

plugins/common/src/modules/elements/image-upload/element.js
bordoni: I think we are ok with the `renderImage`, we can make a move once we are further down on development if we feel the need.

plugins/tickets/src/modules/data/blocks/rsvp/__tests__/selectors.test.js
```diff
@@ -52,4 +52,20 @@ describe( 'RSVP block selectors', () => {
 	it( 'should return the settings open', () => {
 		expect( selectors.getRSVPSettingsOpen( state ) ).toBe( DEFAULT_STATE.settingsOpen );
 	} );
+
+	it( 'should return the header iamge object', () => {
```
mitogh: Typo on `image`

plugins/tickets/src/modules/data/blocks/rsvp/reducer.js
```diff
@@ -76,6 +80,11 @@ export default ( state = DEFAULT_STATE, action ) => {
 				...state,
 				settingsOpen: action.payload.settingsOpen,
 			};
+		case types.SET_RSVP_HEADER_IMAGE:
+			return {
+				...state,
+				headerImage: headerImage( state.headerImage, action ),
```
mitogh: I feel having an extra reducer here to set an object here it might not be required I think the main point is that we are defining the fields of the image, I think we either should the `header` property instead of the specific properties of the object here so either is `null` or is set to an object, in that way we simplify the process to reset this values as well as just setting the value to `null` reset all the fields of the image.  And will remove the need for an extra reducer to handle every single field of the image.

bordoni: @mitogh what do you mean bt extra reducer here? You mean the `headerImage` function call?

mitogh: Not the call it itself, this one specifically: 

- https://github.com/moderntribe/events-gutenberg/blob/59872ad33225d39953764358efc781daac9c23a3/plugins/tickets/src/modules/data/blocks/rsvp/reducers/header-image.js

This one can be avoided as I feel is not really required and just have `SET_RSVP_HEADER_IMAGE` and from there actions can set a single object or `null` it that way things become simpler from setter / getters.

And in order to ensure object has the required signature I think on this one the `PropTypes.shape()` has been implemented so data can be either be `null` or the correct shape. 

Another thing here is having an object with default values for `id`, `src`, and `img` it means a whole object has not been defined which in this case `null` for the entire object seems to communicate that correctly IMHO.

bordoni: I like the idea of having the empty object because we can always count on the three values been there, and we can do the check for `id` as the rule for if it is a valid Image.

mitogh: Right, however: 

1. I think it makes things more complex that it needs to be as `null` on the property indicates the absence of the whole thing. And the selectors should be the ones handling the case if there's no image so components and reducers are unaware of this. 

1. Image object comes already populated and is provided by Gutenberg so if is a non empty object we can rely on the data structure to be accurate.

1. Saving the full image object allows to do selection on different image sizes  and extra properties of the images, this kind of selections can happen at `selector` level.

1. Removes the need to compose reducer as you set either the object response from Gutenberg or `null` to remove it.

bordoni: So you are saying we should use the object Gutenberg sends us for images? And `null` when it doesn't exist?

mitogh: Yes Gutenberg image object looks something like this: 

```js
const image = {
	id: 4961,
	title: 'aircraft-1362586_1920',
	filename: 'aircraft-1362586_1920.jpg',
	url: 'http://gutenberg.local/wp-content/uploads/2018/09/aircraft-1362586_1920.jpg',
	link: 'http://gutenberg.local/event/tickets-here/aircraft-1362586_1920/',
	sizes: {
		full: {
			url: 'http://gutenberg.local/wp-content/uploads/2018/09/aircraft-1362586_1920.jpg',
			height: 1211,
			width: 1920,
			orientation: 'landscape',
		},
		large: {
			height: 331,
			width: 525,
			url: 'http://gutenberg.local/wp-content/uploads/2018/09/aircraft-1362586_1920-1024x646.jpg',
			orientation: 'landscape',
		},
		medium: {
			height: 189,
			width: 300,
			url: 'http://gutenberg.local/wp-content/uploads/2018/09/aircraft-1362586_1920-300x189.jpg',
			orientation: 'landscape',
		},
		thumbnail: {
			height: 150,
			width: 150,
			url: 'http://gutenberg.local/wp-content/uploads/2018/09/aircraft-1362586_1920-150x150.jpg',
			orientation: 'landscape',
		},
	},
};
```

This way the reducer can just accepts the payload either an `object` or `null`

bordoni: Ahh I think for now we can stick to the option, but @paulmskim add on the code a note about not using the full payload from Gutti.

Let's go with the current approach. We can for sure talk about that later on, we will have the chance.

PR #327, opened 2018-09-19
Render the google map in the event venue block.

juanfra:
🎫 https://central.tri.be/issues/113159


plugins/events/src/modules/blocks/event-venue/template.js
```diff
@@ -277,6 +285,26 @@ class EventVenue extends Component {
 			return editVenue;
 		}
 	}
+
+	// Get the coordinates according to the venue address
+	// So we can display the map on the backend
+	getCoordinates = ( details )  => {
```
bordoni: We need to save these coordinates into their correct Meta fields so we dont do a geocode on every load of the page.

PR #328, opened 2018-09-20
Fix/inactive state layout

paulmskim:


mitogh:
Just resolve the conflicts here so we can merge after that.

PR #329, opened 2018-09-20
Add TODO anotation to keep track of central question

mitogh:


plugins/events/src/modules/blocks/event-organizer/template.js
```diff
@@ -131,6 +131,8 @@ class EventOrganizer extends Component {
 
 		removeOrganizerInBlock( clientId, organizer );
 
+		// TODO: make sure this one is provided by the container in mapDispatchToProps, as both
```
bordoni: Actually, lets use
```
/**
 * @todo 
 */
```

PR #332, opened 2018-09-24
RSVP: Use the `post_id` from the main template

juanfra:
🎫 https://central.tri.be/issues/112191

Been checking the views for RSVP on the new architecture. The only problem I found was that the post_id wasn't making it to the guten tickets template class. I think this will be probably resolved later once we make pages and posts compatible with tickets and guten as standalone.

By now, I think this will make us have the functionality in events.

### DIFF
--- a/plugins/tickets/src/Tribe/Blocks/Rsvp.php
+++ b/plugins/tickets/src/Tribe/Blocks/Rsvp.php
@@ -53,7 +53,7 @@ extends Tribe__Gutenberg__Common__Blocks__Abstract {
 	 * @return string
 	 */
 	public function render( $attributes = array() ) {
-		$post_id            = tribe( 'gutenberg.tickets.template' )->get( 'post_id' );
+		$post_id            = tribe( 'gutenberg.events.template' )->get( 'post_id' );
 		$args['attributes'] = $this->attributes( $attributes );
 		$args['tickets']    = $this->get_tickets( $post_id );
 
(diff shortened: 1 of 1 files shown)


plugins/tickets/src/Tribe/Blocks/Rsvp.php
```diff
@@ -53,7 +53,7 @@ public function default_attributes() {
 	 * @return string
 	 */
 	public function render( $attributes = array() ) {
-		$post_id            = tribe( 'gutenberg.tickets.template' )->get( 'post_id' );
+		$post_id            = tribe( 'gutenberg.events.template' )->get( 'post_id' );
```
mitogh: I think TEC is not required in order to have Tickets or Tickets+ running, so doing this, at least for now will do so.

juanfra: I don't think ET +/or ET+ are prepared to run as standalone with this extension right now. That's what I meant with my comment. That once we do that (make tickets + gutenberg work without TEC) we would need to check how we get the post id. Right now the global tpl arguments are set when the main events template is rendered (part of the TEC files of the extension: https://github.com/moderntribe/events-gutenberg/blob/feature/event-ticket-blocks/plugins/events/src/Tribe/Template/Overwrite.php#L82)

mitogh: Right!

However I think adding a `@todo` here so this one gets addressed when the files are placed on each plugin directory as eventually this will be part only of tickets and the other only of TEC. So when that happens we are aware of this one at least at a code level that this one should be updated.

Overall I think we should merge after that so we can unblock folks

juanfra: Thanks buddy!

PR #333, opened 2018-09-25
Fix/113527 timezone display issue

paulmskim:
For ticket: https://central.tri.be/issues/113527


bordoni:
Make sure we have a Changelog.

PR #335, opened 2018-09-26
Fields for RSVP base fields

bordoni:
_Ref:_ [C#112960](http://central.tri.be/issues/112960)
---
This isn't done there are a few fields missing but should allow the creation of RSVP via the correct Rest EndPoint.

`POST` to `/wp-json/v2/wp/tribe_rsvp_attendees`, it's a base WordPress Post Endpoint so creation should follow the standards here: https://developer.wordpress.org/rest-api/reference/posts/#create-a-post

We are still missing going and not going, which I will add tomorrow.

### DIFF
--- a/plugins/tickets/src/Tribe/Meta.php
+++ b/plugins/tickets/src/Tribe/Meta.php
@@ -21,6 +21,12 @@ class Tribe__Gutenberg__Tickets__Meta {
 			$this->text()
 		);
 
+		register_meta(
+			'post',
+			'_stock',
+			$this->text()
+		);
+
 		// Tickets Hander Keys
 		$handler = tribe( 'tickets.handler' );
 
--- a/plugins/tickets/src/Tribe/Meta.php
+++ b/plugins/tickets/src/Tribe/Meta.php
@@ -85,10 +85,6 @@ class Tribe__Gutenberg__Tickets__Meta {
 			$this->text()
 		);
 
-		// $woo = tribe( 'tickets-plus.commerce.woo' );
-		// $edd = tribe( 'tickets-plus.commerce.edd' );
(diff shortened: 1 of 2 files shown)


plugins/tickets/src/Tribe/Meta.php
```diff
@@ -0,0 +1,220 @@
+<?php
+/**
+ * Initialize Gutenberg Event Meta fields
+ *
+ * @since TBD
+ */
+class Tribe__Gutenberg__Tickets__Meta {
```
mitogh: Wonder about this one is only for RSVP or `Tickets` as well because of the naming here mainly

plugins/tickets/src/Tribe/Meta.php
```diff
@@ -0,0 +1,220 @@
...
+			$this->text()
+		);
+
+		register_meta(
+			'post',
+			Tribe__Tickets__Global_Stock::TICKET_STOCK_CAP,
+			$this->text()
+		);
+
+		// $woo = tribe( 'tickets-plus.commerce.woo' );
```
mitogh: I guess this comment shouldn't be part of this PR?

plugins/tickets/src/Tribe/Meta.php
bordoni: `Tribe__Gutenberg__Tickets` is the naming for the Autoloader to find the class, but this class will most likely go away soon, in favor of figuring out a better way to handle the Rest API endpoint meta fields.

So for now I am giving a generic name just like in events

mitogh: Makes sense, thank you!

PR #337, opened 2018-09-27
Buy tickets workflow for different providers.

juanfra:
🎫 https://central.tri.be/issues/115040

📹 **Screencast**

https://www.useloom.com/share/9bd3ebbc4da44cefaeaf07e26d5770e3


plugins/tickets/src/Tribe/Blocks/Tickets.php
```diff
@@ -60,7 +60,8 @@ private function get_tickets( $post_id ) {
 	 * @return string
 	 */
 	public function render( $attributes = array() ) {
-		$post_id            = tribe( 'gutenberg.tickets.template' )->get( 'post_id' );
+		// @todo: We'll need to get the post id from ET later, so support it as standalone
+		$post_id            = tribe( 'gutenberg.events.template' )->get( 'post_id' );
```
bordoni: We have a `tribe( 'gutenberg.tickets.template' )` class.

plugins/tickets/src/resources/js/views/tickets.js
```diff
@@ -55,6 +56,160 @@ tribe.tickets.block = {
...
+			}
+
+			var new_quantity = parseInt( $this.val(), 10 );
+			new_quantity     = isNaN( new_quantity ) ? 0 : new_quantity;
+
+			if ( new_quantity > 0 ) {
+				$form
+					.find( '[data-ticket-id]:not([data-ticket-id="' + $ticket_id + '"])' )
+					.closest( 'div' )
+					.find( 'input, button' )
```
bordoni: We should avoid so open-ended selectors, maybe we could use classes to reduce these selectors a bunch.

plugins/tickets/src/resources/js/views/tickets.js
```diff
@@ -55,6 +56,160 @@ tribe.tickets.block = {
...
+	/**
+	 * Handle the tickets form submission
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	$( obj.selector.container )
+		.on( 'click',
+		obj.selector.submit,
```
bordoni: Weird indentation.

plugins/tickets/src/resources/js/views/tickets.js
```diff
@@ -55,6 +56,160 @@ tribe.tickets.block = {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	obj.handleSubmitWoo = function() {
+
+		// Set the flag to process WooTickets
+		$( '<input>' ).attr( {
+			type  : 'hidden',
```
bordoni: We should make these into a `<script>` template, so we can modify from PHP and allow more things to be added.

plugins/tickets/src/resources/js/views/tickets.js
```diff
@@ -55,6 +56,160 @@ tribe.tickets.block = {
...
+	 *
+	 * @return bool
+	 */
+	obj.handleSubmitTpp = function() {
+
+		// Reset form action as needed
+		$( obj.selector.container ).attr( 'action', '' );
+
+		// Add required post var
+		$( '<input>' ).attr( {
```
bordoni: We should make these into a `<script>` template, so we can modify from PHP and allow more things to be added.

plugins/tickets/src/views/blocks/tickets.php
```diff
@@ -21,9 +21,12 @@
 	class="<?php echo esc_attr( implode( ' ', $cart_classes ) ); ?>"
 	method="post"
 	enctype='multipart/form-data'
+	data-provider="<?php echo esc_attr( $provider->class_name ); ?>"
 	novalidate
 >
+	<input name="provider" value="<?php echo esc_attr( $provider->class_name ); ?>" type="hidden">
```
bordoni: Add a class, to possibly make it easier to fetch.

plugins/tickets/src/Tribe/Blocks/Tickets.php
juanfra: Yes, but the post ID is not there yet. The post ID is set on the events template class

plugins/tickets/src/views/blocks/tickets.php
juanfra: Easier than the name? This is a hidden input.

plugins/tickets/src/Tribe/Blocks/Tickets.php
bordoni: True.

bordoni: use the 
```
/**
 * @todo
 */
```

format

plugins/tickets/src/views/blocks/tickets.php
bordoni: I should have said, faster in JS. Attr selectors are shitty.

plugins/tickets/src/resources/js/views/tickets.js
juanfra: Have in mind that we're doing it over the `$form`, and the data attribute for the ticket ID is only on each ticket wrapper.

plugins/tickets/src/resources/js/views/tickets.js
juanfra: These are required flags by each vendor. If we move it to the PHP side I think we have the hidden input directly. Do you want me to put that within the templates? I made it this way in order to keep the templates as clean as possible, instead of having hidden inputs that are internal (they put data in $_POST so that the plugin handles the vendor/redirection)

plugins/tickets/src/resources/js/views/tickets.js
juanfra: We have it like that in the entire file, do you want me to set all of them into one line?

plugins/tickets/src/resources/js/views/tickets.js
bordoni: I know, but users can have their own inputs or customizations, and open-ended selectors will break when that happens.

plugins/tickets/src/resources/js/views/tickets.js
bordoni: I wanted on `<script>` containing the `<input>` HTML as text template, so we have only one JS for any number of Commerces, just copy the contents of the correct `<script>` element and prepend into the form.

PR #338, opened 2018-09-27
Save the header and capacity attributes

mitogh:
- Persist the header and capacity attributes of the tickets container block 
- Save the content of the child blocks

This PR requires: https://github.com/moderntribe/events-gutenberg/pull/336


plugins/tickets/src/modules/blocks/tickets/index.js
```diff
@@ -24,8 +26,23 @@ export default {
...
+	attributes: {
+		sharedCapacity: {
+			type: 'string',
+			source: 'meta',
+			meta: '_tribe_ticket_capacity',
+		},
+		header: {
+			type: 'string',
+			source: 'meta',
+			meta: '_tribe_ticket_header',
```
paulmskim: This one already exists as a key in `tickets/data/utils.js`, let's use that for now. The way we handle that will change, but for now best to have the keys in one place.

plugins/tickets/src/modules/blocks/tickets/index.js
```diff
@@ -24,8 +26,23 @@ export default {
 		html: false,
 	},
 
-	attributes: {},
+	attributes: {
+		sharedCapacity: {
+			type: 'string',
+			source: 'meta',
+			meta: '_tribe_ticket_capacity',
```
paulmskim: Let's add this key to `tickets/data/utils.js`, much like the ticket header image below.

plugins/tickets/src/modules/blocks/tickets/template.js
```diff
@@ -45,15 +53,21 @@ class TicketsTemplate extends PureComponent {
 		this.updateIsSelected();
 	}
 
-	render() {
+	renderSpinner() {
+		return (
+			<Fragment>
```
paulmskim: no need for `Fragment` here, we're returning only one element.

plugins/tickets/src/modules/blocks/tickets/index.js
mitogh: Makes sense, I looked this on the RSVP PR and I think is a good idea to store all meta keys in constants will update.

mitogh:
Perfect I've addressed those comments, let me know if you have any additional one.

Thanks

PR #339, opened 2018-09-28
Feature/rsvp store data

paulmskim:
Finishing off tickets:

- https://central.tri.be/issues/112940
- https://central.tri.be/issues/112945

I had to keep the REST API calls in thunks given the way the `wpRequest` action was set up. Will move over to sagas once we can better abstract out `wpRequest`.


plugins/tickets/src/modules/blocks/rsvp/action-buttons/attendees-action-button/container.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Internal dependencies
+ */
+import AttendeesActionButton from './template';
+import { selectors } from '@moderntribe/tickets/data/blocks/rsvp';
+import { withStore } from '@moderntribe/common/hoc';
+
+const mapStateToProps = ( state ) => ( {
+	isDisabled: selectors.getRSVPLoading( state ),
```
mitogh: Maybe `getRSVPIsLoading` ?

plugins/tickets/src/modules/blocks/rsvp/action-buttons/settings-action-button/container.js
```diff
@@ -8,14 +8,18 @@ import { compose } from 'redux';
  * Internal dependencies
  */
 import SettingsActionButton from './template';
-import { actions } from '@moderntribe/tickets/data/blocks/rsvp';
+import { actions, selectors } from '@moderntribe/tickets/data/blocks/rsvp';
 import { withStore } from '@moderntribe/common/hoc';
 
+const mapStateToProps = ( state ) => ( {
+	isDisabled: selectors.getRSVPLoading( state ),
```
mitogh: Same as above.

plugins/tickets/src/modules/blocks/rsvp/action-dashboard/container.js
```diff
@@ -4,63 +4,101 @@
...
+
 /**
  * Internal dependencies
  */
 import RSVPActionDashboard from './template';
-import { actions, selectors } from '@moderntribe/tickets/data/blocks/rsvp';
+import { actions, selectors, thunks } from '@moderntribe/tickets/data/blocks/rsvp';
 import { withStore } from '@moderntribe/common/hoc';
 
+const getIsCancelDisabled = ( state ) => (
```
mitogh: I think this one and the next might be a good case to set a selector instead of creating a local variable here.

plugins/tickets/src/modules/blocks/rsvp/container.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+import RSVP from './template';
+import {
+	actions,
+	selectors,
+	thunks
+} from '@moderntribe/tickets/data/blocks/rsvp';
+import { withStore, withSaveData } from '@moderntribe/common/hoc';
+
+const getDateTimeMoment = ( date, time ) => {
```
mitogh: Maybe worth to move into `utils/moment`

plugins/tickets/src/modules/data/blocks/rsvp/reducer.js
```diff
@@ -158,6 +165,11 @@ export default ( state = DEFAULT_STATE, action ) => {
 				...state,
 				hasChanges: action.payload.hasChanges,
 			};
+		case types.SET_RSVP_LOADING:
+			return {
+				...state,
+				loading: action.payload.loading,
```
mitogh: `isLoading` maybe?

plugins/tickets/src/modules/data/utils.js
```diff
@@ -1 +1,14 @@
...
+/* @todo: these are expected to change based on BE changes */
+export const KEY_RSVP_FOR_EVENT = '_tribe_rsvp_for_event';
+export const KEY_TICKET_SHOW_DESCRIPTION = '_tribe_ticket_show_description';
+export const KEY_PRICE = '_price';
+export const KEY_TICKET_CAPACITY = '_tribe_ticket_capacity';
+export const KEY_TICKET_START_DATE = '_ticket_start_date';
+export const KEY_TICKET_END_DATE = '_ticket_end_date';
+export const KEY_TICKET_SHOW_NOT_GOING = '_tribe_ticket_show_not_going';
+
+export const KEY_TICKET_HEADER = '_tribe_ticket_header';
```
mitogh: We might need to consolidate this later as is the same value used on tickets and RSVP maybe move this into a parent level into the reducer so both blocks update / select the value from the same location instead of a block level.

bordoni: We will move it away from here later on, I asked this to be here. 
@paulmskim can you add a `@todo` here?

```
/**
 * @todo
 */
```

plugins/tickets/src/modules/blocks/rsvp/action-dashboard/container.js
paulmskim: I see your point here, I feel that these 2 uses of the combination of selectors would only be used here. I think the broader question is what do we define as selectors and where do we draw the line? Is it to extend reference to the state or is it to cover any and all logic related to the state? At this point, I've drawn the line at extending reference to the state, but happy to discuss this as we move forward.

paulmskim:
I think I've addressed everything, @mitogh @bordoni  let me know if there's anything else.

PR #343, opened 2018-10-02
Feature/persist rsvp header image

paulmskim:
Ticket to persist rsvp header image. Things done here:

* Obtained event post meta, set initial state if post meta exists
* Fetch image data based on image id during set initial state
* Create thunk to get image based on image id
* Create thunks to update post meta on set image and remove image


plugins/tickets/src/modules/blocks/rsvp/header-image/container.js
```diff
@@ -4,14 +4,22 @@
...
 /**
  * Internal dependencies
  */
 import RSVPHeaderImage from './template';
-import { actions, selectors } from '@moderntribe/tickets/data/blocks/rsvp';
-import { DEFAULT_STATE } from '@moderntribe/tickets/data/blocks/rsvp/reducers/header-image';
+import { selectors, thunks } from '@moderntribe/tickets/data/blocks/rsvp';
 import { withStore } from '@moderntribe/common/hoc';
 
+const mapSelectToProps = ( select ) => ( {
```
mitogh: I don't think we need to define the variable `postId` as a property seems like is not used by the template as you can call `select( 'core/editor' ).getCurrentPostId()` anywhere in the code on `mapDispatchToProps` for instance

paulmskim: where would I get the `select` function from? It's not meant to be used in the template, it's used in the `onSelect()` method in `mapDispatchToProps`.

mitogh: woops I overlook just see `map` 👍

PR #344, opened 2018-10-02
Feature/rsvp going not going react

paulmskim:
Things done in this PR:

* Create store components for setting going and not going counts
* Grab going and not going counts from post meta
* Connect state to `Counters` component

This PR requires https://github.com/moderntribe/events-gutenberg/pull/343 to be merged first.


plugins/tickets/src/modules/data/blocks/rsvp/reducers/temp-details.js
```diff
@@ -0,0 +1,80 @@
...
+ */
+import moment from 'moment/moment';
+
+/**
+ * Internal dependencies
+ */
+import { types } from '@moderntribe/tickets/data/blocks/rsvp';
+import { moment as momentUtil } from '@moderntribe/common/utils';
+
+export const DEFAULT_STATE = {
```
mitogh: I think this one should be imported from `details` reducer instead as it should be a copy of the `details` anyway so any change on the `details` reducers only takes place in a single place.

paulmskim: I'm not too picky here, but doing this makes temp details a bit difficult to follow. You're right in that the state structure for `details` and `tempDetails` is exactly the same, but if someone is looking through the code, they would expect the `DEFAULT_STATE` to be within the `temp-details.js` reducer. I will make this change though, as it's not too difficult.

PR #350, opened 2018-10-03
Send and ensure shared capacity does not overflow

mitogh:


plugins/tickets/src/modules/blocks/ticket/edit-container/content/capacity/container.js
```diff
@@ -25,19 +25,43 @@ const mapDispatchToProps = ( dispatch, ownProps ) => ( {
...
 } );
 
+const mergeProps = ( stateProps, dispatchProps, ownProps ) => {
+	return {
+		...stateProps,
+		...dispatchProps,
+		onCapacityChange( value ) {
+			dispatchProps.setCapacity( stateProps.type, stateProps.totalSharedCapacity, value );
+		},
+		...ownProps,
```
paulmskim: `mergeProps()` default return is:
```
{
  ...ownProps,
  ...stateProps,
  ...dispatchProps,
}
```
we should follow that pattern, then add extra props below that. So this just involves moving `...ownProps` above `...stateProps`.

plugins/tickets/src/modules/data/blocks/ticket/sagas.js
```diff
@@ -22,6 +22,7 @@ import {
 } from '@moderntribe/tickets/data/blocks/ticket/reducers/ticket';
 import { wpREST } from '@moderntribe/common/utils/api';
 import { config, restNonce } from '@moderntribe/common/src/modules/utils/globals';
+import { TYPES } from '@moderntribe/tickets/blocks/ticket/edit-container/content/capacity/template';
```
paulmskim: maybe a better place for ticket types to live is in `data/utils.js` or `data/ticket/utils.js`, so that accessing types is from a place that makes sense (probably also call it `TICKET_TYPES` too if you're moving it to utils`)

PR #357, opened 2018-10-04
Removes `_edd_button_behavior` key from REST API

mitogh:
The attribute can't be set via the tickets API and is not used so
removing it from the Tickets pages allows to save the ticket.


mitogh:
Done thanks.

PR #360, opened 2018-10-05
Prevent ticket creations if no provider is present

mitogh:


plugins/tickets/src/modules/data/blocks/ticket/selectors.js
```diff
@@ -307,3 +308,13 @@ export const getTicketAvailability = createSelector(
 	[ getTicketBlock ],
 	( block ) => parseInt( block.available, 10 ) || 0,
 );
+
+export const getProviders = () => {
```
paulmskim: shouldn't this be done in the config and have the providers set in the `setInitialState()` function?

mitogh: However this value does not comes from the `attributes` and is not stored into any reducer and is used to compose some selectors as is a value that is used in some components via those selectors.

PR #369, opened 2018-10-08
Feature/113380 multi day checkbox

paulmskim:
For ticket: https://central.tri.be/issues/113380

Requires https://github.com/moderntribe/events-gutenberg/pull/368 to be merged first.


mitogh:
Do you think you can merge master into recurrence branch to make the diff smaller and a bit more easier to visualize, thanks 🙏

plugins/common/src/modules/elements/checkbox-input/element.js
```diff
@@ -0,0 +1,35 @@
...
+ * Internal dependencies
+ */
+import Input from '@moderntribe/common/elements/input/element';
+import './style.pcss';
+
+const CheckboxInput = ( {
+	checked,
+	className,
+	onChange,
+	...props
```
nealfennimore: Thoughts on changing this to `rest` instead of props? Maybe restProps?

paulmskim: sure thing, no preference to me on what it's called.

plugins/events-pro/src/modules/elements/multi-day-checkbox/__tests__/element.test.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+import { MultiDayCheckbox } from '@moderntribe/events-pro/elements';
+
+describe( 'Multi Day Checkbox Element', () => {
+	it( 'renders multi day checkbox', () => {
+		const component = renderer.create(
+			<MultiDayCheckbox
+				checked={ true }
+				className="test-class"
+				onChange={ noop }
```
mitogh: Not a blocker but it might be better to use `jest.fn` in this context (test) so you can call it later if that one was called otherwise there's no need to import `lodash` by doing so.

plugins/events-pro/src/modules/elements/multi-day-checkbox/element.js
```diff
@@ -0,0 +1,44 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal Dependencies
+ */
+import Checkbox from '@moderntribe/common/elements/checkbox/element';
+
+class MultiDayCheckbox extends PureComponent {
+	static propTypes = {
+		checked: PropTypes.bool.isRequired,
```
mitogh: Not a blocker either but I think we discuss this few PR's ago, when we mentioned the idea to consolidate the boolean names into something like `isChecked` and `isDisabled` across any wrappers.

PR #371, opened 2018-10-08
Make sure object properties match object destruct operation

mitogh:
Ensures the names of the properties match the ones used by the destruct
operation.


mitogh:
Branch name contains the ticket number with the fix for this:

- https://central.tri.be/issues/112467/

In terms of code update here

`dispatchProps.setMultiDay` destructs values  such as: 

```js
export const setMultiDay = ( { start, end, isMultiDay } ) => ( dispatch ) => {
```

Sending a value like `checked` into the param list means `isMultiDay` value destructed by `setMultiDay` is `undefined` same for #372

PR #372, opened 2018-10-08
Prevent reset of times for an event when set from time picker

mitogh:
This PR requires #371

### DIFF
--- a/plugins/events/src/modules/blocks/event-datetime/container.js
+++ b/plugins/events/src/modules/blocks/event-datetime/container.js
@@ -98,14 +98,14 @@ const onStartTimePickerChange = ( stateProps, dispatchProps ) => ( e ) => {
 	setStartTime( { start, seconds } );
 };
 
-const onStartTimePickerClick = ( stateProps, dispatchProps ) => ( value, onClose ) => {
+const onStartTimePickerClick = ( stateProps, dispatchProps ) => ( seconds, onClose ) => {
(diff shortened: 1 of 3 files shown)


mitogh:
Same answer: 

- https://github.com/moderntribe/events-gutenberg/pull/371#issuecomment-427933203

Just different destruction of keys name.

plugins/events/src/modules/blocks/event-datetime/container.js
```diff
@@ -98,14 +98,14 @@ const onStartTimePickerChange = ( stateProps, dispatchProps ) => ( e ) => {
 	setStartTime( { start, seconds } );
 };
 
-const onStartTimePickerClick = ( stateProps, dispatchProps ) => ( value, onClose ) => {
+const onStartTimePickerClick = ( stateProps, dispatchProps ) => ( seconds, onClose ) => {
 	const { start, end } = stateProps;
 	const { setStartTime, setAllDay } = dispatchProps;
 
-	const isAllDay = value === 'all-day';
+	const isAllDay = seconds === 'all-day';
```
paulmskim: any way we can keep this as `value` and pass in `seconds: value` below? just thinking it doesn't make sense with `seconds === 'all-day'`, as the value could be a number or `all-day`. what do you think?

mitogh: I think that will make things more clear 👍

PR #373, opened 2018-10-09
Feature/recurrence series end

paulmskim:
For ticket: https://central.tri.be/issues/115753

Creates series ends component

Depends on https://github.com/moderntribe/events-gutenberg/pull/369

PR #374, opened 2018-10-09
Feature/115758 day of week picker

paulmskim:
For ticket: https://central.tri.be/issues/115758

Depends on https://github.com/moderntribe/events-gutenberg/pull/373

### DIFF
--- a/plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
+++ b/plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import React from 'react';
-import { noop } from 'lodash';
 
 /**
  * Internal dependencies
@@ -21,13 +20,13 @@ describe( 'Day Of Week Picker Element', () => {
 				thursdayChecked={ true }
 				fridayChecked={ false }
 				saturdayChecked={ true }
-				onSundayChange={ noop }
-				onMondayChange={ noop }
-				onTuesdayChange={ noop }
-				onWednesdayChange={ noop }
-				onThursdayChange={ noop }
-				onFridayChange={ noop }
-				onSaturdayChange={ noop }
+				onSundayChange={ jest.fn() }
+				onMondayChange={ jest.fn() }
+				onTuesdayChange={ jest.fn() }
+				onWednesdayChange={ jest.fn() }
+				onThursdayChange={ jest.fn() }
+				onFridayChange={ jest.fn() }
+				onSaturdayChange={ jest.fn() }
 			/>
 		);
 		expect( component.toJSON() ).toMatchSnapshot();
--- a/plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
+++ b/plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
@@ -1,0 +1,35 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import { noop } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import DayOfWeekPicker from '../element';
+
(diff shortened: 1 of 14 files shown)


plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
```diff
@@ -0,0 +1,35 @@
...
+			<DayOfWeekPicker
+				className="test-class"
+				sundayChecked={ true }
+				mondayChecked={ false }
+				tuesdayChecked={ true }
+				wednesdayChecked={ false }
+				thursdayChecked={ true }
+				fridayChecked={ false }
+				saturdayChecked={ true }
+				onSundayChange={ noop }
```
mitogh: `jest.fn`

plugins/events-pro/src/modules/elements/on-day-of-week/element.js
```diff
@@ -0,0 +1,67 @@
...
+	<Row className={ classNames(
+		'tribe-editor__events-pro__row--on-day-of-week',
+		'tribe-editor__on-day-of-week',
+		className
+	) }>
+		<Label className="tribe-editor__on-day-of-week__label">
+			{ __( 'On', 'events-gutenberg' ) }
+		</Label>
+		<div className="tribe-editor__on-day-of-week__content">
+			<DayOfWeekPicker
```
mitogh: Curios on design of this one as seems like a bit similar on props reused. An array of objects representing each might seems like an alternative.

paulmskim: Currently I don't know what the data structure is going to be like. If the data structure ends up being something easier to manage as an object or array, I think I can update it then.

plugins/events-pro/src/modules/elements/day-of-week-picker/__tests__/element.test.js
```diff
@@ -0,0 +1,34 @@
...
+			<DayOfWeekPicker
+				className="test-class"
+				sundayChecked={ true }
+				mondayChecked={ false }
+				tuesdayChecked={ true }
+				wednesdayChecked={ false }
+				thursdayChecked={ true }
+				fridayChecked={ false }
+				saturdayChecked={ true }
+				onSundayChange={ jest.fn() }
```
nealfennimore: Are there specific cases where we'll need to know when certain days have changed?

plugins/events-pro/src/modules/elements/day-of-week-picker/element.js
```diff
@@ -0,0 +1,103 @@
...
+const DayOfWeekPicker = ( {
+	className,
+	sundayChecked,
+	mondayChecked,
+	tuesdayChecked,
+	wednesdayChecked,
+	thursdayChecked,
+	fridayChecked,
+	saturdayChecked,
+	onSundayChange,
```
nealfennimore: Are there specific cases where we'll need to know when certain days have changed? Could this be abstracted out into a single change handler?

paulmskim: No specific case, I suppose we can pass in values when it changes to know which day has changed. I started with `day of week`, then linearly scaled up so that's what this has become. But I'm sure once the state is fleshed out this will simplify significantly. I could abstract out now if you think that's the way to go.

nealfennimore: No worries, I was just curious to see if there was a reason. I already see it reusing the same change handler in the `on-day-of-week` element. It's good to go :shipit:

paulmskim: Will definitely consolidate at the lower level if structure makes sense. The single `onChange` handler is a placeholder for now, since the details aren't fully drawn up.

PR #376, opened 2018-10-09
Feature/115755 single to date time picker

paulmskim:
For ticket: https://central.tri.be/issues/115755

Depends on https://github.com/moderntribe/events-gutenberg/pull/375

PR #378, opened 2018-10-09
Feature/115756 from time range picker

paulmskim:
For ticket: https://central.tri.be/issues/115756

Depends on: https://github.com/moderntribe/events-gutenberg/pull/376


paulmskim:
@nealfennimore as I was working through this I wondered if the `Row` component could be more structured rather than just taking in `children`, that way we could use the CSS just once on the structured `Row` component. Thoughts?

PR #394, opened 2018-10-16
Ticket/116164 merge store

mitogh:


plugins/common/src/modules/data/plugins/constants.js
```diff
@@ -0,0 +1,2 @@
+export const EVENTS_PLUGIN = 'events';
+export const EVENTS_PRO_PLUGIN = 'events-pro';
```
paulmskim: Should `Tickets` and `Tickets Plus` be defined here? Or should `Events Pro` move to the `events-pro` plugins folder?

plugins/events-pro/src/modules/data/index.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * External dependencies
+ */
+import reducer from './reducers';
+
+import { actions, constants } from '@moderntribe/common/data/plugins';
+import { store } from '@moderntribe/common/store';
+const { EVENTS_PRO_PLUGIN } = constants;
```
paulmskim: Regarding the comment above, if we're only using this constant in the `events-pro` plugin maybe we should define it in this plugin instead.

plugins/common/src/modules/data/plugins/constants.js
nealfennimore: `events-pro` should stay here, we need it for injecting blocks into other plugins, and it removes the load dependency issues that would come from keeping it in other plugin directories.

Tickets and Tickets Plus should be here as well

mitogh: Most of the code of this branch was cherry-pick from `recurrence` branch but it was not committed as such to avoid the merge of files non used on this branch like recurrence elements.

On the other hand yes all plugins should be placed here and tickets has a `TODO` on his data code.

- https://github.com/moderntribe/events-gutenberg/blob/master/plugins/tickets/src/modules/data/index.js#L11

Was not added as this changes are still not in `master` so as soon this branch or `recurrence` is in `master` we should be able to add `tickets`  here as well.

PR #398, opened 2018-10-16
Add Preview component

mitogh:
Create a `<Preview>` component to render the output of an Additional Field on how it should look when the edit was completed.


plugins/events-pro/src/modules/blocks/additional-fields/index.js
```diff
@@ -31,7 +33,12 @@ export default {
 
 	attributes: {},
 	edit: () => (
+		<div>
```
nealfennimore: Can we use `Fragment` here?

plugins/common/src/modules/elements/heading/element.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Internal dependencies
+ */
+import './style.pcss';
+
+const Heading = ( { level, children, className } ) => {
+	const HeadingLevel = `h${ level }`;
+	const headingClassName = classNames(
+		'tribe-editor__heading',
+		`tribe-editor__h${ level }`,
```
paulmskim: This should be `tribe-editor__heading--h${ level }` instead, as it's a modification of the heading element.

plugins/common/src/modules/elements/heading/style.pcss
```diff
@@ -0,0 +1,16 @@
+.tribe-editor__heading {
+	font-family: 'Helvetica', 'Arial', 'sans-serif';
+	color: #000;
+}
+
+.tribe-editor__h1 {
```
paulmskim: These should be fixed as well with the change above.

plugins/common/src/modules/elements/paragraph/element.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+import './style.pcss';
+
+export const SIZES = {
+	medium: 'medium',
+	small: 'small',
+};
+
+const Paragraph = ( { children, size, className } ) => (
+	<div
```
paulmskim: If this element is a paragraph element, shouldn't we be using a `<p>` tag? Otherwise we should just be using a `<div>` element, no?

mitogh:
Changes has been addressed

mitogh:
Would you mind take another look @paulmskim thanks

PR #403, opened 2018-10-17
Add <Settings> component

mitogh:
Create general componet used to control additional fields, with option for `after` and `before` to extend component for Checkbox Behavior.

**Ticket** https://central.tri.be/issues/96627


plugins/events-pro/src/modules/blocks/additional-fields/elements/settings/container.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+import SettingsTemplate from './template';
+import { withStore } from '@moderntribe/common/hoc';
+import { config } from '@moderntribe/common/utils/globals';
+
+/**
+ * @todo get data from a selector
+ */
+const getSettingsLink = () => {
+	const eventsPro = config().events_pro || {};
```
paulmskim: This works to pass in the config value to the component, but we'll need to figure out how to handle these `config()` calls better, they're not really a state variable (unless we are storing them in the state, but in this case I can't see the need for that) and this isn't really what `connect()` is meant for. I see the `todo` above, but we'll need to chat about this later when we pass in more data from php.

plugins/events-pro/src/modules/blocks/additional-fields/elements/settings/template.js
```diff
@@ -0,0 +1,43 @@
...
+import { InspectorControls } from '@wordpress/editor';
+
+const Settings = ( { name, before, after, settingsLink } ) => (
+	<InspectorControls key="inspector">
+		<PanelBody title={ sprintf( __( '%1$s Settings', 'events-gutenberg' ), name ) }>
+			{ before }
+			{ ! ! settingsLink && (
+				<span>
+				{ __( 'Adjust this block’s options under Events → Settings → ', 'events-gutenberg' ) }
+					<a
```
paulmskim: indentation

plugins/events-pro/src/modules/blocks/additional-fields/elements/settings/template.js
```diff
@@ -0,0 +1,43 @@
...
+import { sprintf } from 'sprintf-js';
+
+/**
+ * Wordpress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { PanelBody } from '@wordpress/components';
+import { InspectorControls } from '@wordpress/editor';
+
+const Settings = ( { name, before, after, settingsLink } ) => (
```
paulmskim: Do you think we'll use this in other aspects of the codebase, say in the `events` plugin? Should we move this to `common` instead?

plugins/events-pro/src/modules/blocks/additional-fields/elements/settings/container.js
mitogh: Right, I think we need to consolidate this, and in order to keep components like templates outside of this type of things I decided to create a container that injects this properties instead.

I think a possible solution to inject global variables into components might be the use of a HOC similar to connect but specifically to inject variables from global state into the template.

plugins/events-pro/src/modules/blocks/additional-fields/elements/settings/template.js
mitogh: I think we can abstract this a bit more in order to do so as this specifc settings has a `PanelBody` in the middle that contains a link for the additional fields on it, which is a common pattern for all the different types of additional fields so in order to make it more universal I think we might need to change that section of the `PanelBody` into a param so it can be set as a parameter instead

mitogh:
Would you mind take another look @nealfennimore 

Thanks

PR #405, opened 2018-10-18
Make sure the ticket block uses the currency from the ticket provider

juanfra:
🎫 https://central.tri.be/issues/115649

Make sure the currency is respecting the ticket provider currency, both when it's loaded from the rest API and when they want to create a new ticket.

These changes require to have this one merged: https://github.com/moderntribe/event-tickets/pull/889


plugins/tickets/src/modules/data/blocks/ticket/reducers/__tests__/__snapshots__/ticket.test.js.snap
```diff
@@ -128,6 +138,7 @@ Object {
   "available": 0,
   "capacity": "",
   "capacityType": "own",
+  "currencySymbol": undefined,
```
mitogh: I see some `undefined` here I guess it might be a string or something along those lines.

plugins/tickets/src/modules/data/utils.js
```diff
@@ -30,3 +32,32 @@ export const TICKET_ORDERS_PAGE_SLUG = {
...
+export function getProviderCurrency( provider ) {
+	const tickets = config().tickets || {};
+	const providers = tickets.providers || {};
+
+	// if we don't get the provider, return the default one
+	if ( '' == provider ) {
+		return tickets.default_currency;
+	}
+
+	const index = providers.findIndex( el => el.class === provider );
```
mitogh: I don't think we have a polyfill for this one `findIndex` it might be better to do it with `filter` instead.

bordoni: Good catch

plugins/tickets/src/modules/data/blocks/ticket/reducers/__tests__/__snapshots__/ticket.test.js.snap
bordoni: Null would be the best

juanfra: Thanks guys!

The snapshot was updated automatically, I'm not super familiar with them @mitogh, should I update that manually?

juanfra:
Ok, I'm not sure about the tests one. Let me know.

PR #407, opened 2018-10-18
Add shape for additional fields reducers in store

mitogh:
Create:

- Selectors
- Types
- Actions
- Reducers

For additional fields

Add a blocks folder following same `recurrence` structure to avoid possible merge conflicts.

**Ticket** https://central.tri.be/issues/114552


plugins/events-pro/src/modules/data/blocks/additional-fields/reducers/fields.js
```diff
@@ -0,0 +1,50 @@
...
+	}
+};
+
+const allFields = ( state = [], action ) => {
+	switch ( action.type ) {
+		case types.ADD_ADDITIONAL_FIELD:
+		case types.SET_ADDITIONAL_FIELD_NAME:
+			return [ ...state, action.payload.name ];
+		case types.REMOVE_ADDITIONAL_FIELD:
+			return [ ...state ].filter( ( name ) => name !== action.payload.name );
```
nealfennimore: The spread for state is redundant here - filter will always return a new array :)

PR #413, opened 2018-10-19
Feature/116023 connect recurrence data 2

nealfennimore:
- Moves over logic to sagas
- Adds in syncRule action to prevent edit actions from looping

### DIFF
--- a/plugins/events-pro/src/modules/data/blocks/recurring/selectors.js
+++ b/plugins/events-pro/src/modules/data/blocks/recurring/selectors.js
@@ -44,7 +44,7 @@ export const getStartDate = createSelector(
 
 export const getStartTime = createSelector(
 	[ getRule ],
-	( rule ) => rule.start_time.slice( 0, -3 ),
+	( rule ) => rule.start_time
 );
 
 export const getEndDate = createSelector(
@@ -54,7 +54,7 @@ export const getEndDate = createSelector(
 
 export const getEndTime = createSelector(
 	[ getRule ],
-	( rule ) => rule.end_time.slice( 0, -3 ),
+	( rule ) => rule.end_time
 );
 
 export const getBetween = createSelector(
--- a/plugins/events-pro/src/modules/elements/exception-field/monthly.js
+++ b/plugins/events-pro/src/modules/elements/exception-field/monthly.js
@@ -16,6 +16,7 @@ import {
 	RecurringToDateTimePicker,
 	SeriesEnds,
 } from '@moderntribe/events-pro/elements';
+import { constants } from '@moderntribe/events-pro/data/blocks';
 import { options } from '@moderntribe/events-pro/data/blocks/exception';
 
 const MonthlyField = ( { isMultiDay, typeOption, onTypeChange } ) => {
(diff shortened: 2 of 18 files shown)


plugins/events-pro/src/modules/elements/from-time-range-picker/element.js
```diff
@@ -21,94 +21,37 @@ const {
...
-	const { startTime } = stateProps;
-	const min = toSeconds( startTime, TIME_FORMAT_HH_MM );
-
-	if ( toSeconds( e.target.value, TIME_FORMAT_HH_MM ) <= min ) {
-		return;
-	}
 
-	dispatchProps.editRule( ownProps.index, { end_time: `${ e.target.value }:00` } );
+const onMultiDayChange = ( stateProps, dispatchProps, ownProps ) => ( e ) => {
+	dispatchProps.editRule( ownProps.index, { multi_day: !! e.target.value } );
```
nealfennimore: We should really be using the next constant keys here

plugins/events-pro/src/modules/data/blocks/recurring/selectors.js
```diff
@@ -44,7 +44,7 @@ export const getStartDate = createSelector(
 
 export const getStartTime = createSelector(
 	[ getRule ],
-	( rule ) => rule.start_time.slice( 0, -3 ),
+	( rule ) => rule.start_time
```
nealfennimore: @paulmskim 	this was causing errors so I took it out. There's some inconsistencies between the format used here and the time pickers

plugins/events-pro/src/modules/elements/exception-field/monthly.js
```diff
@@ -26,11 +27,15 @@ const MonthlyField = ( { isMultiDay, typeOption, onTypeChange } ) => {
 				options={ options.EXCEPTION_OCCURRENCE_OPTIONS }
 				selected={ typeOption }
 				onChange={ onTypeChange }
+				blockType={ constants.EXCEPTION }
 			/>
 			<DayOfMonthPicker />
 			<FromTimeRangePicker />
 			{ isMultiDay && <RecurringToDateTimePicker /> }
-			<SeriesEnds rowLabel={ __( 'Exception ends', 'events-gutenberg' ) } />
+			<SeriesEnds
```
paulmskim: this and all the `SeriesEnds` elements should also get an `index` prop

plugins/events-pro/src/modules/elements/from-time-range-picker/element.js
```diff
@@ -21,94 +21,37 @@ const {
...
-	if ( isMultiDay ) {
-		const max = toSeconds( endTime, TIME_FORMAT_HH_MM );
+const onEndTimeChange = ( stateProps, dispatchProps, ownProps ) => ( e ) => {
+	const seconds = toSeconds( e.target.value, TIME_FORMAT_HH_MM );
+	const min = toSeconds( stateProps.startTime, TIME_FORMAT_HH_MM );
 
-		if ( toSeconds( e.target.value, TIME_FORMAT_HH_MM ) >= max ) {
-			return;
-		}
+	if ( seconds > min ) {
```
paulmskim: This check should only happen if `multiDay` is `false`. So we have a few cases:
- If `multiDay = false` and if `seconds <= min`, then return, we won't update time
- If `multiDay = false` and if `seconds > min`, then update time.
- If `multiDay = true`, then update time, as there are no constraints.
We can do this in sagas, but we need to handle this logic somewhere.

plugins/events-pro/src/modules/elements/from-time-range-picker/element.js
```diff
@@ -21,94 +21,37 @@ const {
...
+		dispatchProps.editRule( ownProps.index, { end_time: e.target.value } );
```
paulmskim: The `TimePicker` input returns a time in `HH:MM` format, but the state needs to hold the time in `HH:MM:SS` format. We need to update this somewhere, whether it's sagas or here.

plugins/events-pro/src/modules/data/blocks/recurring/selectors.js
```diff
@@ -44,7 +44,7 @@ export const getStartDate = createSelector(
 
 export const getStartTime = createSelector(
 	[ getRule ],
-	( rule ) => rule.start_time.slice( 0, -3 ),
```
paulmskim: We need to either do this here or do it when we map state to props, as we're storing the time as `HH:MM:SS` format in the state but using it as `HH:MM` format in the `TimePicker` component.

PR #411, opened 2018-10-19
Abstract Field component for multiple field types

mitogh:
The PR introduces a single Component called `Field` that provides the option to render any other Field with the same layout used on all the additional fields design.

The following PR requires the next PRs to be merged:

- https://github.com/moderntribe/events-gutenberg/pull/409
- https://github.com/moderntribe/events-gutenberg/pull/404
- https://github.com/moderntribe/events-gutenberg/pull/403
- https://github.com/moderntribe/events-gutenberg/pull/407

PR #416, opened 2018-10-23
Render components based on types

mitogh:
- Render different types of components based on the type of field
- Connect fields with the store
- Connect properties with the block attributes
- Add sagas for custom fields


plugins/events-pro/src/modules/data/blocks/additional-fields/actions.js
```diff
@@ -3,6 +3,18 @@
  */
 import { types } from '@moderntribe/events-pro/data/blocks/additional-fields';
 
+export const setInitialState = ( props ) => ( {
+	type: types.SET_ADDITIONAL_FIELD_INITIAL_STATE,
+	payload: props,
+} );
+
+export const setChange = ( name ) => ( {
```
paulmskim: not a blocker here, but it would be nice to try and keep things consistent from `type` to `actions` to change handlers to `sagas`, etc. It makes the code cleaner and much easier to follow as you know what to expect and know what to search for.

For example, if I know the type is `SET_ADDITIONAL_FIELD_CHANGE`, I would expect the action to be called `setAdditionalFieldChange()`, but here it's called `setChange()`. In this case, it's not what I expected, and I have to jump around looking for it. Same thing with the sagas, where we have the watched action type called `SET_ADDITIONAL_FIELD_CHANGE` but the generator called `setPristineState()`. If we know the naming conventions and one of the names, we can extrapolate the rest and jump through the code much easier than having to confirm the name of each action, change handler, saga, etc.

plugins/events-pro/src/modules/blocks/additional-fields/checkbox/template.js
```diff
@@ -9,49 +9,46 @@ import PropTypes from 'prop-types';
...
-};
+const CheckboxInput = ( { options, onChange } ) => (
+	<fieldset className="tribe-editor__additional-fields__edit--horizontal-fields">
+		{ options.map( ( option, index ) => {
+			const { label = '', value = '', isChecked = false } = option;
+			const name = normalize( label );
+			const id = `name-${ index + 1 }`;
+			return (
+				<Fragment key={ id }>
+					<input
```
paulmskim: As mentioned in https://github.com/moderntribe/events-gutenberg/pull/411 could we try and use the `Checkbox` element here? Or is there a good reason why we aren't using the existing element?

plugins/events-pro/src/modules/blocks/additional-fields/radio/template.js
```diff
@@ -10,44 +10,41 @@ import PropTypes from 'prop-types';
...
+const RadioInput = ( { options, onChange, selectedValue } ) => (
+	<fieldset className="tribe-editor__additional-fields__edit--horizontal-fields">
+		{ options.map( ( option, index ) => {
+			const { label = '', value = '' } = option;
+			const name = normalize( label );
+			const id = `name-${ index + 1 }`;
+			const isChecked = value === selectedValue;
+			return (
+				<Fragment key={ id }>
+					<input
```
paulmskim: Same here as the `Checkbox` element, could we create a `Radio` element just like the `Checkbox` element we have in `common` and use that? Also applies to the text input in `additional-fields/text/template.js` and the url input in `additional-fields/url/template.js`.

plugins/events-pro/src/modules/data/blocks/additional-fields/actions.js
mitogh: I do see the inconsistency here, however I think there a couple of additional things to consider as well.

- As the `exports` happens to a file level is clear where the function is coming from and adding an extra prefix just increases verbosity here into something like: 

```js
export { setAdditionalFieldChange } from '@moderntribe/events-pro/data/blocks/additional-fields'
```

- Another thing is usually this actions are used only inside of the same `additional-fields` 

- In terms of the type they all are under a single namespace so they might have a higher chance to collapse if we don't set a prefix on it so I think types should have the prefix on it.

I think this might be a good topic to discuss as well with other folks thoughts on this but above are mine.

Let me know what do you think, thanks @paulmskim

mitogh:
💃 @nealfennimore

plugins/common/src/modules/elements/textarea/element.js
```diff
@@ -0,0 +1,21 @@
...
+ * Internal dependencies
+ */
+
+const Textarea = ( { className, ...rest } ) => (
+	<textarea className={ classNames( 'tribe-editor__textarea', className ) } { ...rest } />
+);
+
+Textarea.propTypes = {
+	className: PropTypes.string,
+	type: PropTypes.string.isRequired,
```
paulmskim: may want to remove this

plugins/common/src/modules/elements/url-input/element.js
```diff
@@ -0,0 +1,28 @@
...
+import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import Input from '@moderntribe/common/elements/input/element';
+
+const UrlInput = ( { checked, className, onChange, ...rest } ) => (
+	<Input
+		checked={ checked }
```
paulmskim: no need for checked here

plugins/events-pro/src/modules/blocks/additional-fields/dropdown/template.js
```diff
@@ -10,21 +10,21 @@ import PropTypes from 'prop-types';
...
 
-const DropdownField = ( { name, value, options, onInputChange, ...rest } ) => (
+const DropdownField = ( { name, value, options, onInputChange, isSelected, output } ) => (
 	<Field
-		name={ name }
-		input={ <Select options={ options } value={ value } onChange={ onInputChange } /> }
-		output={ value }
-		{ ...rest }
+		id={ name }
+		input={ <Select options={ options } value={ value } onChange={ onInputChange } isSearchable={ false } /> }
```
paulmskim: You'll also want to add `backspaceRemovesValue={ false }` here, as currently when you have a selected value, you can remove it by hitting `backspace`.

plugins/events-pro/src/modules/data/blocks/additional-fields/actions.js
paulmskim: Definitely, I understand this standpoint. If it wasn't obvious I'm more of the camp of being more verbose to make things clearer and easier to follow, but definitely something to discuss and align on a clear direction for how we write our JS.

PR #417, opened 2018-10-23
Register additional fields into the WP Rest API

mitogh:
- Create an `abstract` class to keep `Meta` files DRY
- Create an interface to keep a contract between parent / child classes
- Update `Meta` on TEC and Tickets classes to remove repated code
- Add Meta class for PRO to include dynamic additional fields


mitogh:
Sure I think I can add this in this PR @juanfra however I'm not sure which one do you mean specifically, would you mind point to an example.

Thanks 👍

juanfra:
Sure! For example:

Here: https://github.com/moderntribe/events-gutenberg/blob/a78766056f013ba9ea13f52c3239e9e941a9752b/plugins/common/src/Tribe/Meta.php#L20

Here: https://github.com/moderntribe/events-gutenberg/blob/a78766056f013ba9ea13f52c3239e9e941a9752b/plugins/common/src/Tribe/Meta.php#L36

Here: https://github.com/moderntribe/events-gutenberg/blob/a78766056f013ba9ea13f52c3239e9e941a9752b/plugins/common/src/Tribe/Meta.php#L53

mitogh:
I did an update so would you mind take another look @juanfra i'm still unsure if that's are the requested changes 😅 , thanks.

PR #420, opened 2018-10-24
Save data from store into the post

mitogh:
Use the attributes of the block to store the data from attributes into the meta fields when the fields are saved, so additional fields changes are persisted

**Ticket**

https://central.tri.be/issues/114552

**Dependencies**

- https://github.com/moderntribe/events-gutenberg/pull/417


plugins/events-pro/src/modules/data/blocks/additional-fields/selectors.js
```diff
@@ -100,16 +100,21 @@ export const getFieldOptionsWithLabels = createSelector(
 
 export const getFieldDropdownValue = createSelector(
 	[ getFieldBlock ],
-	( field ) => field.value || {},
+	( field ) => ( { value: field.value, label: field.value } ),
```
nealfennimore: Label is intended to be the same as the value? Just double checking

mitogh: Yes, as the DB only stores a single value, we don't have a `label`, however the  `Select` component uses an object to construct every single value.

So the label and value are the same because the value is displayed as was entered in the settings so which means display the value as is stored in the DB .

nealfennimore: Cool just noting in case it wasn't :)

mitogh: Thanks for asking 😄

PR #421, opened 2018-10-24
Register the Additional Fields templates

mitogh:
Add the additional fields templates after the `tribe/event-venue` block for
new events

**Ticket**

- https://central.tri.be/issues/114527

**Requires**

-  https://github.com/moderntribe/events-gutenberg/pull/417


plugins/common/src/Tribe/Meta.php
```diff
@@ -0,0 +1,148 @@
...
+	}
+	
+	/***
+	 * Default definition for an attribute of type boolean
+	 *
+	 * @since 0.2.4-alpha
+	 *
+	 * @return array
+	 */
+	protected function boolean() {
```
bordoni: Let’s pick either protected or private

mitogh: Hey I'm not sure about this one do you mean move `protected` into `private` or the other way around?

PR #422, opened 2018-10-25
Feature/store connection

nealfennimore:
Allows for saving recurrence rules and exclusions into DB as JSON string and then parsing it back out to use with UI


plugins/events-pro/src/modules/blocks/recurrence-exception/template.js
```diff
@@ -12,37 +12,45 @@ import ExceptionForm from '@moderntribe/events-pro/elements/exception-form/eleme
...
-						<ExceptionForm
-							exceptions={ this.props.exceptions }
-							removeException={ this.props.removeException }
-							editException={ this.props.editException }
-						/>
-						<ExceptionAddField onClick={ this.props.addField } noBorder />
-					</Panel>
-				)
-				: <ExceptionAddField onClick={ this.props.initialExceptionPanelClick } />
+			<Fragment>
```
paulmskim: wondering why you need the `<Fragment>`

plugins/events-pro/src/modules/blocks/recurrence-rule/template.js
```diff
@@ -15,31 +15,38 @@ import Panel from '@moderntribe/events-pro/elements/panel/element';
...
-					>
-						<RecurringForm
-							rules={ this.props.rules }
-							removeRule={ this.props.removeRule }
-						/>
-						<RecurringAddField onClick={ this.props.addField } noBorder />
-					</Panel>
-				)
-				: <RecurringAddField onClick={ this.props.initialRulePanelClick } />
+			<Fragment>
```
paulmskim: Same question here

plugins/events-pro/src/modules/elements/attribute-sync/template.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * External dependencies
+ */
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+export default class AttributeSync extends Component {
```
paulmskim: This is an interesting way of doing this sync, is there possibly another way of doing this? Also, this would be a great case for passing in data via a global, but as we're not there yet, I suppose we're going to have to get creative.

plugins/events-pro/src/modules/blocks/recurrence-exception/template.js
nealfennimore: Ah I don't, it was removed in a recent commit. I didn't remove the Fragment with it

plugins/events-pro/src/modules/elements/attribute-sync/template.js
nealfennimore: I'd love a different way of doing this, but there's some issues:
- We're stuck using `setAttributes` and `clientId` as they'r'e passed down via props, and it's tricky to get that action dispatched on to the `core/editor` store from within our sagas. This component is the interface for passing that method/id in for reuse.
- Since `props.attributes.rules` is a JSON string, we need a way of putting it in the store as an array of objects. And vice versa, we need to take the store array of objects and convert them back to a JSON string via `setAttributes` upon saving. Sagas in this case keep everything synced together, so we don't have to manage this (but in reality we should really only be deriving state from the attributes)

I don't like this interface, it's not as clean as I'd like. Gutenberg introduces a weird API when it comes to sharing, as we cannot from what I can tell, get the attributes directly via a selector, but rather we always have to pass it down via the root component of the block.

I'm not sure what you mean by a global, can you give an example?

paulmskim: Yeah, like we've been discussing before, if we could spit out all the data into a window object and rehydrate the store before we render the components, we wouldn't need something like this. But as it is, that is not set up at the moment.

PR #424, opened 2018-10-25
Task/fix elements

paulmskim:


plugins/events-pro/src/modules/elements/series-ends/element.js
```diff
@@ -45,12 +52,28 @@ const onSeriesEndsChange = ( ownProps, edit ) => ( selectedOption ) => (
...
 
-const onSeriesEndsOnDateChange = ( ownProps, edit ) => ( date ) => {
-	const endDate = date
-		? momentUtil.toDate( momentUtil.toMoment( date ), dateUtil.FORMATS.DATABASE.datetime )
-		: '';
-	edit( ownProps.index, { [ KEY_END_DATE ]: endDate } );
-};
+const onSeriesEndsOnDateChange = ( ownProps, edit, end ) => (
+	( date, modifiers, dayPickerInput ) => {
+		console.log('here in day change');
```
nealfennimore: :video_game:

plugins/events-pro/src/modules/elements/series-ends/template.js
```diff
@@ -35,6 +35,8 @@ const SeriesEnds = ( {
 
 		if ( seriesEnds && seriesEnds.value === constants.DATE ) {
 			const seriesEndsOnDateObj = new Date( seriesEndsOnDate );
+			console.log( 'seriesEndsOnDate', seriesEndsOnDate );
```
nealfennimore: :video_game:

plugins/events-pro/src/modules/data/blocks/recurring/selectors.js
```diff
@@ -25,27 +48,37 @@ export const getRule = createSelector(
 
 export const getType = createSelector(
 	[ getRule ],
-	( rule ) => rule.type,
+	( rule ) => rule[ KEY_TYPE ],
```
nealfennimore: I'll make a note to do this with exception selectors. Probably will move the selectors to shared folder and then import into here for easier reuse

paulmskim: yeah like 95% of these selectors are the same, so would be nice to get them using the same, just pass in an exception selector or recurrence selector.

PR #426, opened 2018-10-26
Feature/fix elements 2

nealfennimore:
- Share selectors between exception/rules
- Connect multi-day to work properly
- Add in multi_day_span for multi-day events ending within a week


plugins/events-pro/src/modules/elements/recurring-form/element.js
```diff
@@ -33,6 +33,7 @@ export default class RecurringForm extends PureComponent {
 							isMultiDay={ rule[ KEY_MULTI_DAY ] }
 							onRemoveClick={ this.props.removeRule }
 							type={ rule[ KEY_TYPE ] }
+							{ ...rule }
```
paulmskim: just wondering why we need to pass in the rule here if we have the index already.

nealfennimore: alright it's out now :+1:

PR #434, opened 2018-10-27
Feature/116989 app subdirectory

nealfennimore:
https://central.tri.be/issues/116989

- Moves react assets to app subdir
- Updates to export both min and unmin CSS + JS files on production builds


borkweb:
Does `plugins/package.json` need to be updated so that the `main` property reads `src/resources/app/main.js` rather than `src/resources/main.js`?

nealfennimore:
No it shouldn't - it's not really used in this project

nealfennimore:
@bordoni go ahead and merge this when you're ready

PR #440, opened 2018-10-30
Escape boostrap script

mitogh:


plugins/events-pro/scripts/linkDependencies
```diff
@@ -1,5 +1,5 @@
 #! /bin/bash
 
-ln -sf $(pwd)/src/modules/blocks $(pwd)
-ln -sf $(pwd)/src/modules/data $(pwd)
-ln -sf $(pwd)/src/modules/elements $(pwd)
+ln -sf "$(pwd)"/src/modules/blocks "$(pwd)"
```
nealfennimore: Let's change it to `"$(pwd)/src/modules/blocks"`

nealfennimore: Go ahead and merge it afterwards

PR #445, opened 2018-10-30
Fix/116907 timepicker styling disabled

nealfennimore:
https://central.tri.be/issues/116907


plugins/common/src/modules/elements/time-picker/style.pcss
```diff
@@ -12,6 +12,8 @@
 	justify-content: center;
 
 	input[type="time"] {
+		font-family: "Noto Serif",serif;
```
paulmskim: space after comma?

PR #447, opened 2018-10-31
updating block icons

zbtirrell:
See: https://central.tri.be/issues/117020

We also updated a couple of the block names per Leah

screenshot 2018-10-30 23 56 51


zbtirrell:
snapshots are fixed, tests are now passing. (at some point I should learn how to run these locally…)

Crisoforo - I totally agree on the mouseovers and such, but let's handle that as additional polish in another pass.

mitogh:
Thank you.

Sound good. You can execute tests locally by running `npm run test` just make sure all dependencies has been installed `npm run install`

PR #451, opened 2018-11-01
Fix/115804 capacity ui

nealfennimore:
https://central.tri.be/issues/115804

Fixes shared tickets interactions
- Allows shared capacity to only be set on initial shared ticket creation. Subsequent shared tickets have optional limit.
- Sets shared optional limit to shared limit by default
- Fixes available and capacity counts

https://www.useloom.com/share/2d63576b7fd94827b62949bf74b3a95f


mitogh:
Looks great a change log entry would be fantastic.

nealfennimore:
Ugh I keep forgetting to do that

PR #454, opened 2018-11-02
Feature/117277 remove events dependencies

nealfennimore:
https://central.tri.be/issues/117277

Removes events dependencies within common folder.
- Moves withDetails hoc to events
- Moves forms reducer and withForm HOC to common


paulmskim:
Just some failing tests here.

PR #472, opened 2018-11-06
Feature/117227 ticket ecommerce options

paulmskim:
For ticket: https://central.tri.be/issues/117227


plugins/tickets/src/modules/data/blocks/ticket/__tests__/constants.test.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Internal dependencies
+ */
+import { constants } from '@moderntribe/tickets/data/blocks/ticket';
```
nealfennimore: Let's just iterate through the constants next time in a loop and make them snapshots. It'll make our lives a lot easier for maintaining

PR #473, opened 2018-11-06
Feature/common externals

nealfennimore:
This re-enables common externals and allows all plugin bundles to be properly exported on the `window.tribe` object.

Also redefines load order from PHP land to load each individual JS bundle in the correct order.


plugins/events-pro/src/Tribe/Assets.php
```diff
@@ -39,34 +39,34 @@ public function register() {
...
 		tribe_asset(
 			$plugin,
-			'tribe-pro-gutenberg-elements',
-			'elements.js',
-			/**
-			 * @todo revise this dependencies
-			 */
-			array( 'react', 'react-dom', 'wp-components', 'wp-api', 'wp-api-request', 'wp-blocks', 'wp-i18n', 'wp-element' ),
+			'tribe-pro-gutenberg-blocks-styles',
+			'blocks.css',
```
bordoni: `app` directory?

PR #475, opened 2018-11-06
add conditional to update natural language label in saga

paulmskim:
For ticket: https://central.tri.be/issues/117235


plugins/events/src/modules/data/blocks/datetime/sagas.js
```diff
@@ -32,7 +32,10 @@ export function* setHumanReadableLabel( dates = {} ) {
 	}
 
 	const updatedLabel = yield call( rangeToNaturalLanguage, dates.start, dates.end );
-	yield put( actions.setNaturalLanguageLabel( updatedLabel ) );
+
+	if ( currentLabel !== updatedLabel ) {
```
faction23: Nice catch Paul

PR #477, opened 2018-11-06
Include Theme lookup for front-end facing Templates

### LINKED ISSUE #476: Can't override template in theme
With this plugin active, my single-events.php does nothing. We need to make some changes to the banner and h1 positioning to match the rest of the theme. With it being coded into the single-event.php in this plugin, we can't do anything. It seems the single-event.php in this plugin is the supreme overriding template, rather than the one in our theme.

bordoni:
Fixes #476


mitogh:
Merge conflicts

PR #482, opened 2018-11-07
Fix/tribe js config

nealfennimore:
- Removes duplicate blocks.js asset loading
- Adds `tribe_js_config` to be localized before the google maps script. Updates google maps script to load before common assets

### DIFF
--- a/plugins/events/src/Tribe/Editor.php
+++ b/plugins/events/src/Tribe/Editor.php
@@ -345,6 +345,8 @@ class Tribe__Gutenberg__Events__Editor extends Tribe__Gutenberg__Common__Editor 
 			),
 		);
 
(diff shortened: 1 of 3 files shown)


plugins/events/src/Tribe/Editor.php
```diff
@@ -463,29 +467,8 @@ public function assets() {
...
-			'enqueue_block_editor_assets',
-			array(
-				'in_footer' => false,
-				'localize'  => array(),
-				'conditionals' => array( $this, 'is_events_post_type' ),
-				'priority'  => 106,
-			)
-		);
 
 		$localize_blocks = array(
```
paulmskim: @nealfennimore shouldn't we move these into the priority 1 enqueue?

nealfennimore: Oh yes let me do that

PR #484, opened 2018-11-07
Add Queue status `admin_ajax` response

mitogh:
Make an `admin_ajax` response available to keep track and trigger the Queue on events without leaving the page, specifically for large series of recurrence events, ideally a response to this should be every 0.5 to keep consistency with old UI if the series hasn't been completed.

Usage example: 

```js
const data = new FormData();
data.append('action', 'gutenberg_events_pro_recurrence_queue');
data.append('recurrence_queue_status_nonce', tribe_js_config.rest.nonce.queue_status_nonce );
data.append('post_id',  wp.data.select( 'core/editor' ).getCurrentPostId() );

const params = {
  method: 'POST',
  credentials: 'same-origin',
  body: data,
};

// Sample request.
fetch( ajaxurl, params )
.then( (r) => r.json() )
.then( (c) => console.log( c ) );
````

Be aware that: 

- `ajaxurl` is a global localized by WP
- `tribe_js_config.rest.nonce.queue_status_nonce` is a new localized variable with this PR.
- `wpSelect( 'core/editor' ).getCurrentPostId()` function provided by Gutenberg to get the current post ID.

A `false` is returned if the response fails or if the event is non a recurrent event, otherwise an associative array is returned with the following shape: 

```js
{
    completed: true / false
    percentage: [0, 100] 
}
```

This one is available as an Object after parsing into a JSON object.

### DIFF
--- a/plugins/events-pro/src/Tribe/Provider.php
+++ b/plugins/events-pro/src/Tribe/Provider.php
@@ -22,6 +22,7 @@ class Tribe__Gutenberg__Events_Pro__Provider extends tad_DI52_ServiceProvider {
 
 		$this->container->singleton( 'gutenberg.events-pro.meta', 'Tribe__Gutenberg__Events_Pro__Meta' );
 		$this->container->singleton( 'gutenberg.events-pro.recurrence.provider', 'Tribe__Gutenberg__Events_Pro__Recurrence__Provider' );
+		$this->container->singleton( 'gutenberg.events-pro.recurrence.queue-status', 'Tribe__Gutenberg__Events_Pro__Recurrence__Queue_Status' );
 
 		$this->hook();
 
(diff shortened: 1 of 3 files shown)


plugins/events-pro/src/Tribe/Recurrence/Queue_Status.php
```diff
@@ -0,0 +1,97 @@
...
+	
+	/**
+	 * Exit and return the response as json encoded string
+	 *
+	 * @since TBD
+	 *
+	 * @param $data
+	 */
+	public function response( $data ) {
+		exit( json_encode( $data ) );
```
juanfra: Wouldn't it be better to return and then exit? that way the method could be testable, right?

mitogh: Didn't think on that way but I 💯 agree with that will update it.

nealfennimore:
Just putting this on hold for a bit. I'm testing out the responses when there's no series being made. I want to make sure it's usable when the response is just `false`

mitogh:
Okay be aware that the ajax call before this PR will return a `400` as the action is not present on the admin.

nealfennimore:
Yup working off your branch now :+1:

nealfennimore:
@mitogh okay good to merge :cake:

plugins/events-pro/src/Tribe/Provider.php
```diff
@@ -39,6 +40,7 @@ protected function hook() {
 		// Initialize the correct Singletons
 		tribe( 'gutenberg.events-pro.assets' );
 		tribe( 'gutenberg.events-pro.recurrence.provider' )->hook();
+		tribe(  'gutenberg.events-pro.recurrence.queue-status' )->hook();
```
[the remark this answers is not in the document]
mitogh: Fixing on the next PR thanks.

PR #487, opened 2018-11-07
Feature/117770 block ui editing

nealfennimore:
https://central.tri.be/issues/117770

Disables editing of event datetime block when series is being made in the background


plugins/events-pro/src/modules/data/status/sagas.js
```diff
@@ -0,0 +1,127 @@
...
+	try {
+		const payload = new FormData();
+		const postId = wpSelect( 'core/editor' ).getCurrentPostId();
+
+		if ( ! postId ) {
+			throw 'No post ID';
+		}
+
+		yield call( [ payload, 'append' ], 'action', 'gutenberg_events_pro_recurrence_queue' );
+		yield call( [ payload, 'append' ], 'recurrence_queue_status_nonce', window.tribe_js_config.rest.nonce.queue_status_nonce ); // eslint-disable-line max-len
```
faction23: I don't know if there's some constraint in this system but could we have like a globals (or something) file somewhere? It would safely drill into these step by step and export parts so we don't have to this all over the place, while also not erroring if the object doesn't make it out correctly from php land.

PR #489, opened 2018-11-08
Remove recurrence keys if `_EventRecurrence` is deleted

mitogh:
Remove the keys associated with the event:

- `_tribe_blocks_recurrence_exclusions`
- `_tribe_blocks_recurrence_rules`

If the event removes the `_EventRecurrence` meta key as is no longer a 
recurrent event.

Removes the recurrence meta box from the editor as well if
`classic-editor` is set.

### DIFF
--- a/plugins/events-pro/src/Tribe/Meta.php
+++ b/plugins/events-pro/src/Tribe/Meta.php
@@ -14,8 +14,10 @@ class Tribe__Gutenberg__Events_Pro__Meta {
 	 * @return void
 	 */
 	public function register() {
-		register_meta( 'post', '_tribe_blocks_recurrence_rules', tribe( 'gutenberg.events.meta' )->text() );
-		register_meta( 'post', '_tribe_blocks_recurrence_exclusions', tribe( 'gutenberg.events.meta' )->text() );
+		/** @var Tribe__Gutenberg__Events_Pro__Recurrence__Blocks_Meta $blocks_meta */
+		$blocks_meta = tribe( 'gutenberg.events-pro.recurrence.blocks-meta' );
+		register_meta( 'post', $blocks_meta->get_rules_key(), tribe( 'gutenberg.events.meta' )->text() );
+		register_meta( 'post', $blocks_meta->get_exclusions_key(), tribe( 'gutenberg.events.meta' )->text() );
 		
 		add_filter( 'get_post_metadata', array( $this, 'filter_going_fields' ), 15, 4 );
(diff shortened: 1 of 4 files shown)


plugins/events-pro/src/Tribe/Meta.php
```diff
@@ -83,4 +85,37 @@ public function get_value( $post_id = 0, $meta_key = '' ) {
...
+	 * removed.
+	 *
+	 * @since TBD
+	 *
+	 * @param $meta_id
+	 * @param $object_id
+	 * @param $meta_key
+	 */
+	public function remove_recurrence_meta( $meta_id, $object_id, $meta_key ) {
+		if ( '_EventRecurrence' !== $meta_key ) {
```
juanfra: Are we storing these meta keys in any class variable? Perhaps it's better to use those if it's defined anywhere else on PRO

plugins/events-pro/src/Tribe/Meta.php
```diff
@@ -83,4 +85,37 @@ public function get_value( $post_id = 0, $meta_key = '' ) {
...
+			return;
+		}
+		delete_post_meta( $object_id, '_tribe_blocks_recurrence_rules' );
```
juanfra: The same for these two deletes

plugins/events-pro/src/Tribe/Meta.php
bordoni: Not this is not been stored unfortunatelly.

plugins/events-pro/src/Tribe/Meta.php
bordoni: But these two I would like to see us storing.

plugins/events-pro/src/Tribe/Meta.php
```diff
@@ -83,4 +85,37 @@ public function get_value( $post_id = 0, $meta_key = '' ) {
...
+	 * Remove the recurrence meta box if classic-editor is set
+	 *
+	 * @since TBD
+	 *
+	 * @param $show_meta
+	 *
+	 * @return bool
+	 */
+	public function remove_recurrence_classic_meta( $show_meta ) {
+		$is_classic_editor = tribe_get_request_var('classic-editor', null );
```
bordoni: Spacing after `(`

plugins/events-pro/src/Tribe/Meta.php
mitogh: 😞

mitogh:
Changes has been applied guys @bordoni @juanfra – Thanks for taking a look, let me know what do you think.

PR #499, opened 2018-11-09
Add number of items created and date of the last one

mitogh:
Send into the ajax response the number of items created by the
recurrence rule, also returns the date of the last item created


mitogh:
- 'items_created' => has a number with the number of instances created.
- 'last_created_at' => string if has items `false` if '0` is the number of items above.

PR #503, opened 2018-11-09
Notification/events created

nealfennimore:
Adds in progress notice showing how many recurring events have been made and up to what point  https://d.pr/free/i/uMAJIf


mitogh:
Seems like some tests are failing.

